PR #33, opened 2021-12-27
Fraction of area within historic district

SashaWeinstein:
Have code to calculate "Area within a historic district" indicator that calculates both total area of historic districts within each PUMA and fraction of each PUMA that is a historic district. Total area is returned in units of square miles. My code returns a pandas dataframe and doesn't have any way to cache the results. 

I get PUMA geographies from https://www1.nyc.gov/site/planning/data-maps/open-data/census-download-metadata.page

I use the EPSG 2263 projection based on this [blog post](https://gis.stackexchange.com/questions/149291/equidistant-cartesian-projection-for-nyc)

There are some visualizations of overlays between historic districts and PUMAs in the bottom of the `Area_within_Historic_District_Exploratory` notebook.


SashaWeinstein:
Accidentally pointed merge to `main` instead of `dev`

PR #83, opened 2022-01-11
Three or more maintenance deficiencies indicator partially completed

SashaWeinstein:
# Third EDDT pull request 🌱  

This pull request is code that can generate some data points for the units with three or more maintenance deficiencies indicator. This code only aggregates on borough, not PUMA. The race crosstab  isn't implemented yet. I chose to open a pull request before the indicator was complete as a lot of new code was added to clean the housing and vacancy survey (HVS) data. Some refactoring to move the race assign to the root of the `/aggregate` folder was also done. 

### Add new indicator
Code to create new indicator is in `aggregate/housing_security/three_or_more_maintenance_deficiencies.py`. Processing was pretty straightforward and code to use replicate weights to produce standard error on counts was already in place. 

### Improve HVS ingestion process

- `ingest/HVS/HVS_geography_clean.py` was added to convert from borough from numeric codes to strings. This was easy but converting the boro + sub-borough area to PUMA will take some work. That code should live in this file
- `ingest/clean_replicate_weights.py` standardizes replicate weights to reduce logic that has to live in aggregation step
- `ingest/HVS/HVS_ingestion.py` calls functions in files above and also sets weights to correct units. 

### Refactor PUMS aggregation
- `aggregate/PUMS/aggregate_PUMS.py` had two pieces of logic moved to the root of `/aggregate`
- `aggregate/race_assign.py` was created to hold logic on assigning race. It's important that all indicators get the same race categories and so this logic should live in a single file referenced by different aggregation processes 
- `aggregate/clean_aggregated.py` all this does now is put each indicator next to it's standard error. But as we build out the process to export our files I think we will need more functions that standardize outputs. 

### Add tests
The data points the new indicator can produce are tested. As new data points (aggregate by PUMA, crosstab by race) are added to the indicator then more tests should be written.

### Internal review process
`internal_review/` folder was added so that my results can be shared in .csv. No clue if this is best practice as I made it up last minute

### Clean up 
`aggregate/housing_security/housing_security.sh` was deleted as dev container is used instead. This change should not occurred in a different branch

## Conclusion
I know that EDDT is currently on the messy side and the other engineers are putting a lot of effort into understanding the project so far. I'm very happy to take feedback and get a better collaborative process going.


aggregate/housing_security/three_or_more_maintenance_deficiencies.py
```diff
@@ -0,0 +1,51 @@
...
+from aggregate.clean_aggregated import sort_columns
+from ingest.clean_replicate_weights import rw_cols_clean
+from internal_review.set_internal_review_file import set_internal_review_files
+
+implemeted_years = [2017]
+implemented_geographies = ["Borough"]
+
+
+def count_units_three_or_more_deficiencies(
+    geography_level, year, crosstab_by_race=False, requery=False
```
AmandaDoyle, line 15: Do you mention needing to layer in the race ethnicity breakdown during standup?

SashaWeinstein: Yes I believe I mentioned it. I definitely meant to

td928 (approved):
all look good to me.

PR #101, opened 2022-01-14
86 access transit

SashaWeinstein:
Two new indicators in QOL added. These are easiest indicators possible, data is already aggregated by PUMA. Source data is added to data library via manual upload, there is accompanying data library pull request for that change. Results are in in `/internal_review` folder.
I like unit testing to make sure all PUMAs are present. This is straightforward test that can be run for all indicators


AmandaDoyle (commented):
See wiki page on field conventions to update field formats.
Both indicators are described as a percent of residents, but `pop_with_access_subway_SBS` is a count.  
Rename fields to `per_pop_subway_sbs_access` and `per_pop_ada_subway_access`

SashaWeinstein:
Renamed the fields starting with "fraction" to what Amanda specified above. Also changed `PUMA` column name to `puma` and added in the leading zero.  
I took out the counts entirely, it would be easy for me to add them back in if that would be best

PR #104, opened 2022-01-20
27 historic district area

SashaWeinstein:
Two changes are made to the indicator
The first is that puma, borough columns conform to convention laid out in wiki
Second is that citywide aggregation level was added


mbh329 (approved):
* [ ] area_within_historic_district.py changes look good, adds important helper functions and follows geo naming conventions
* [ ] ingestion_helpers.py, looks good - follows leading 0 example set up in wiki 
* [ ] csv's are in the right format using geography as the index and data as columns
* [ ] set_internal_review_File.py makes sense to me but te and I are using the older version of this function 
* [ ] Geography helpers makes sense, follows geo convetions laid out in readme and renames the puma column to lowercase

When you have time, should just rename gp to gpd

PR #105, opened 2022-01-20
26 change in units

td928:
new pull request for change in units


SashaWeinstein (changes requested):
Hey Te most of this looks good I think you're are pretty close. I have a couple comments
1. dataloading.sh shouldn't have been changed the way it was. The existing downloads should be kept in
2. The output files need a couple changes. They should have exactly one row per geography. Also the boroughs should be the two letter codes and the pumas should have a leading 0. You can see the conventions [here](https://github.com/NYCPlanning/db-equitable-development-tool/wiki/New-data-pipeline:-field-naming-convention)
3. This one is optional, I should have been clearer about how this indicators should be generated. I see that all the code can be run via __name__=="__main__", but it should also be accessed by the python code that will generate the final output files. I think it should be a quick refactor to put everything above line 136 into functions that takes geography level and returns a dataframe.
Hope those directions are clear, happy to talk through what I'm thinking

td928:
hey @SashaWeinstein thanks for the notes they are very helpful. The new commits should address the majority of them.

SashaWeinstein (commented):
This looks good! The last thing is the refactor so that there is one function that returns the results. Right now the code to collate the external review files would need to call `load_housing_data` and then pass the `df` and `census10` variables to `units_change_citywide` or `units_change_borough`. We don't want the external review code to do all that. A simple wrapper function will be sufficient

PR #106, opened 2022-01-20
30 housing ny

mbh329:
Merge in affordable housing csvs and python script for housing production indicators : affordable housing construction and preservation


SashaWeinstein (changes requested):
This looks great, just want to request two changes
1. the `sys.path.append(os.path.join(os.path.dirname(__file__),'../../'))` should be removed, the code should only be run from the root directory of the repo
2. boroughs should be represented as codes and pumas should have leading 0 as outlined in [the wiki](https://github.com/NYCPlanning/db-equitable-development-tool/wiki/New-data-pipeline:-field-naming-convention)
This all looks great, think we are real close

SashaWeinstein (approved):
Awesome, this looks good. go ahead and merge in

PR #107, opened 2022-01-21
External review housing production

td928:
still need to wait for Max changes merged in to create the final combined external review files. Script ready to be reviewed.


SashaWeinstein (changes requested):
This is a good start, but there are some issues
The first is that the merge just adds columns together, but it doesn't do a join. You can see that the brooklyn area historic is on the same row as the manhattan change in units. pandas's merge has parameters to tell it what to merge on. My preferred approach is setting the columns to join on as the index and then passing `left_index` and `right_index` as true
I also don't see where in the external review code one can specify the geography level. I think the best way to approach is to write a function that takes geography level as an argument and does a join on just that geography level

td928:
Hey @SashaWeinstein  I think I was able to side step the join issue by setting the dataframe index to the geographies and then do pd.concat. I think it should handle the general case. Please let me know if this makes sense and happy to talk through if needed.

SashaWeinstein (changes requested):
Looks like the merge worked! Great job. I would not have written the code that way but we can talk design later, this is definitely good enough to hit our deadline
I eyeball checked the historic district numbers against the numbers I produced for the internal review so the merge is definitely good

Can you undo giving the files  `.csv.csv` extensions? Then I think it should be good to go.

mbh329:
This looks good to me as well - I think we should try to settle on same naming conventions for our csv's similar to the way we write issues. happy to chat about this whenever

td928:
Hey @SashaWeinstein could you point to the file I made .csv.csv mistake. I don't think the final merged final has an extra csv extension

td928:
ok got it. I think this doesn't impact this work for now. Should we merge this in?

SashaWeinstein:
Ok I think I figured out the issue, it's the historic indicators that are saved as .csv.csv in the internal review folders that these read from. Can you rename these and then we merge?
Also @mbh329 can you do an eyeball check that the numbers for your indicator are the same in the internal review and the external? As an additional set of eyes for quality just to be sure. 
Sorry if this is overkill, we have a couple minutes before standup and I feel like we might as well just use this time to clean and check

SashaWeinstein:
Eyeball checked housing production myself, it all looks good

mbh329:
The data for my indicator looks good, I checked it against a few pumas and boroughs

AmandaDoyle (commented):
- [x] Add underscore "_" between production and geography (i.e. housing_production_borough)
- [x] Update the `geography` column in each csv to be the name of the geography (i.e. puma, borough, city)
- [x] Rename `fraction_area_historic` to `area_historic_pct` and change format to whole digits rounded to two decimal points to meet field naming convention.  
- [x] Rename `total_area_historic` to `area_historic_XYZ` XYZ being the unit sqmi or?

SashaWeinstein:
Uploaded to digital ocean, all looks good

mbh329:
Awesome, looks great

td928:
Haven’t checked digital ocean and but this all look good.

AmandaDoyle (approved):

PR #114, opened 2022-01-27
Add 2008-2012 PUMS

SashaWeinstein:
# Adding the second PUMS survey of 2008-2012
There are two changes to the ingestion logic required to process this new survey.
1. The PUMA information is now found as variable in *either* `PUMA00` (PUMAs associated with 2000 census) or `PUMA10` (PUMAs associated with 2010 census). 
2. The occupation information that was found in the `OCCP` column in the 2015-2019 column is now split across `OCCP02`, `OCCP10`, and `OCCP12`. These columns need to be combined into one

## New PUMA Filter
Because there is no unified PUMA value for each record in this survey, the PUMA information is stored in two columns of `PUMA00` and `PUMA10`. The `PUMA10` column is null for survey responses collected before the 2010 census PUMAs were finalized, and `PUMA00` column null for records collected afterwards.

In the ingestion process for the 2015-2019 survey, filtering for a set PUMAs was done via a single `geoid` for each record. However for this survey each set of PUMAs is associated with two GET requests, one filter on `PUMA00` and the other to filter on `PUMA10`. This doubles the number of GET requests per pull as compared to the 2015-2019 survey, increasing from 6 to 12. 

The files modified to accommodate this change are `PUMS_query_manager.py` and `PUMS_data.py`. New approaches to generate the filters and iterate through the urls were developed. 

## `OCCP` recode
The only change to the actual variables in the 2008-2012 survey is the `OCCP` variable being replaced with `OCCP02`, `OCCP10`, and `OCCP12`. The `Occupation_Variable_2012_PUMS.ipynb` shows my work in examining these new columns and coming to the conclusion that they are mutually exclusive and can be collapsed down to one `OCCP` column. 

This cleaning process was simple but it required a refactor of my process to get variables for a particular "variable type" (demographics or economics). Before it was a simple one to many mapping where demographics and economics mapped to a set of variables that was the same in the URL and the final dataframe. But now the processes to get variables to put in the url and variables to clean are separate. The logic to handle this was put into a new file called `variable_generator.py`.

I'm not totally confident that the codes in `OCCP02`, `OCCP10`, and `OCCP12` mean exactly the same thing as the codes in `OCCP`. I did some exploratory work in the notebook to look at this but it doesn't answer the question. MDAT describes the `OCCP02` as "Occupation recode for data collected in 2009 based on 2002 OCC codes" and says `OCCP10` and `OCCP12` are based on 2010 OCC codes. So we could ask a SME if the 2002 and 2012 OCC codes should be processed differently if no one on DE knows.  

### Update tests
I has to add some new code to the tests to get them to work but they do test the new dataset for things like if all records are unique, if weights are in dataframe, etc.


AmandaDoyle (approved):

mbh329 (approved):
Te and I reviewed logic and ran output on local machine. looks good

PR #124, opened 2022-02-04
change age bucket and testing or age bucket

td928:
## Changes Made

Change the age bucket to reflect the feedbacks from population. Also modified the testing modules to reflect the changes to the updated age buckets.


mbh329 (approved):
This matches the age brackets erica mentioned

PR #125, opened 2022-02-04
New variance measures: margin of error and coefficient of variation

SashaWeinstein:
# Adding margin of error (MOE) and coefficient of variation (CV)

So far all of our data points had have been calculated with only one measure of variance: standard error (SE). This pull request implements processes to replace SE with MOE and add CV.

## How calculations are done

This is a review but it's helpful context. The calculations are done with the R survey package which we call via the rpy2 python package. We use this package to get variance measures from the replicate weights. There are three files in `/statistical` responsible for using the survey package to do calculations: `calculate_counts.py`, `calculate_fractions.py`, `calculate_medians.py`. 

Each of these was modified to return a CV and replace SE with MOE.

### Calculate MOE

The `MOE.py` file in statistical takes a dataframe and calculates the MOE using the a z-score corresponding to the 90% confidence interval, calculated with the python `stats` package. 

It can either replace the SE or keep the SE in. In retrospect this functionality was not worth the time invested to build, should have written it to always replace the SE. 

All calls to any calculate code now includes an `add_MOE` flag and `keep_SE` flag that are to be set by whomever calls the code. `add_MOE` defaults to true and `keep_SE` defaults to false.

### Calculate CV

The `vartype` parameter to the `svyby` function from the survey package can be set to return a CV. In all calls to `svyby` this parameter is set to `["se", "ci", "var", "cv"]` which are 4 variance measures that `svyby` can return. We only need `cv` but it doesn't work when only `cv` is passed. I'm not sure why.

We take CV as a decimal instead of a percentage, it should be easy to convert back to percentage later on if we need to

## Debug fractions crosstabs

The existing code get crosstabs functions backward. When we crosstab by race we want the fraction of people in each category that are each racial group. So for example in labor force we want the fraction of all people in labor force that are white. What we were doing is fraction of all white people in the labor force.

While debugging the code was simplified. The `calculate_fractions_crosstabs` function in `calculate.py` was removed and instead the `calculate_fractions` function takes a single flag `parent_category` that constructs the column label to distinguish whether the fraction is a crosstab.  

## Categories with no respondents

In these cases our counts and fractions should be reported as 0 and the associated measures of variance reported as NA. 

The survey package reports the count of a category with zero respondents as NA. This is converted  to 0 in `counts_to_zero` in `calculate_counts.py`. It reports the variance as NA as well which is correct and doesn't need to be cleaned.

The survey package reports the fraction of a category with zero respondents as 0 which is correct (although not intuitive). It reports the associated SE as 0 however which is not correct. This is converted to NA in `SE_to_zero_no_respondents`

There is also the cleaning of median calculation for these situations. Because the only median we do now is age, the buckets are just PUMA-race and obviously we have survey respondents in each bucket. So there is no corresponding situation to test. However when we do median wages by industry we know there are some industry-race buckets with zero respondents. My decision is to wait until we start that indicator to do this cleaning on medians. I think that adding a dummy category to the existing pipeline is to much effort. 

## Tests

The tests are pretty simple. They were helpful to check that the code runs but were a pain to work with due to long run times. 

Currently each file does it's own aggregation of each data point which takes forever. This could be refactored to use a single aggregated dataframe across multiple tests. 

A couple changes to existing tests were added, maybe they shouldn't be in this pull request.

### Miscellaneous 

The aggregation tests don't run on each push as they never pass and it's not helpful

All calls to calculate code should pass a copy of the dataframe so that operations on the data in the calculate code don't effect the data stored in the class attribute. Good practice generally I think

The `​return f'occupation-{occupation_mapper.get(person["OCCP"], "none")}` line will be made irrelevant when we correctly set the denominator for occupation.


SashaWeinstein:
Added files for internal review. These files should have margin of error, coefficient of variation, and standard error. Our final output won't have standard error but it's helpful for review

mbh329 (approved):
The logic makes sense to me to add the statistical variables to the dataset. My one concern moving forward (I think we will know more after our chat with labs this afternoon) is column naming conventions in the outputs.

SashaWeinstein:
Ok neat. In my opinion we should merge these changes in and then worry about column naming later. @AmandaDoyle is that the right approach?

PR #127, opened 2022-02-09
Fix fraction calculations of PUMS demographic indicators

SashaWeinstein:
# Improvements to PUMS demographic fractions aggregation

I closed my previous pull request and opened a new one to start fresh. 

This pull request includes a set of improvements to the PUMS demographic indicators. It was started to implement a custom denominator but over the course of developing this change more issues were identified and addressed. 

## Add Denominator Column 
This is the place to start as it's a simple improvement that makes the output data more readable. It's something Tyler asked for during our meeting Monday afternoon. This column is set in `calculate_fractions`:
```python
denom = data.groupby(geo_col).sum()["PWGTP"]
single_fraction["denominator"] = denom
```
It's renamed to "indicator-fraction-denom" in the final output table. 
The denominator is set as the total *people* (not survey respondents) in the dataframe passed to `calculate_fractions`. So setting the denominator involves applying a filter to a dataframe before it's passed to `calculate_fractions`. 
This happens in two places, in  [applying a custom denominator](#lep-custom-denominator) and [fixing the race crosstab](#fraction-of-race-group-in-category)

## LEP Custom denominator
The process to apply a custom denominator was already written so all that needed to be added is this function:
```python
def over_five_filter(self, PUMS: pd.DataFrame):
subset = PUMS[PUMS["AGEP"] >= 5]
return subset
```
This function is referenced in the `indicators_denom` attribute of the `PUMSCountDemographics` class. The improvement that was the original intention of the branch took around 5% of the total time. 

## Fraction of race group in category
Previously the fractions reported in race crosstabs were the fraction of people in a particular category in that racial group. So PopU16 black non-hispanic was fraction of people under sixteen who were black. What we should be reporting in that column is the fraction of black non-hispanic people who are under sixteen.  
This change involved changing the `add_fractions` in the `PUMSAggregator` class to iterate through races instead of iterating through categories. So previously it would iterate through the age buckets and pass a subset of all records in that age bucket to be crosstabbed by race. Because subset passed to `calculate_fractions` was people in that category, the denominator was people in that category. By switching to iterating and subsetting by race in the `PUMSAggregator` method the denominator was switched to people in that race. 

## Report data points in affirmative cases only
Previously we were reporting stats those who were NOT foreign born, NOT limited english proficiency, etc. This process was changed to only report the affirmative cases. This reduces the size of the outputted dataset and reduces complexity as we don't need to distinguish between those whose english is good and those who english is N/A. 

## Checking and testing
There is a jupyter notebook `notebooks/LEP_custom_denom.ipynb` that lays out the new columns and how they relate to each other and pytests in `tests/PUMS/test_PUMS_denominator_columns.py`to formally test the results. There is also a .csv in `internal_review/demographics/PUMA` that has the results of aggregating the test data.


mbh329 (approved):
Denominator logic is clear to me. Was able to run the code on my local machine with expected results

PR #129, opened 2022-02-09
117 automate external review

SashaWeinstein:
The good news is that it the workflow works. It automatically combines indicators of housing production into one file per geography. As we change column names or add new indicators it will be easy to collate the files and send them to digital ocean.

The bad news is that according to [the documentation](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) "your workflow must be in the default branch" to trigger using the green button on the github site. So we have to run via pushes which requires changing the category directly in the .yml which isn't as nice.

Here is a link to a successful run https://github.com/NYCPlanning/db-equitable-development-tool/actions/runs/1820182800


mbh329 (approved):
Logic is clear - a lot of the refactoring are related to formatting issues with the older code (make sure to use a python linter in vscode - Bandit is preferred). 

Refactored the geography helper functions and incorporated them across the 3 separate agg scripts. Separate `assing_PUMA.py `from `geography_helpers`

Changed the `export_DO.sh` to take any csv in the proper format and send it to DO 

`external_review_collate` - logic makes sense to me although not familiar with typer module

PR #130, opened 2022-02-09
Add Civilian Filters to Economic PUMS Indicators by Industry 

mbh329:
We want to make sure we are not including any military respondents in the data - add civilian population filter and remove "military" as an industry in `count_PUMS_economics.py`. Also need to make sure that we aren't dropping columns in the join - specify "left join" in  `aggregate_PUMS.py`


SashaWeinstein (commented):
This looks good! Thanks for fixing the pull request text. And thanks for making that fix to the merge.

Ran numbers on my machine. Industry has a custom denominator whereas labor force does not, which is correct. 

Go ahead and merge when you're ready

PR #131, opened 2022-02-10
add empty dictionary Initialization to pums aggregator 

mbh329:


SashaWeinstein (approved):
looks good, this should definitely be here. Good practice to write up a description of the problem that prompted the change

PR #133, opened 2022-02-14
PUMS economics medians - wages by occupation and industry

SashaWeinstein:
# Wages by occupation and industry
There are three indicators that we get medians of in the economic category: wages by occupation, wages by industry, and median household income. This pull request implements code to do wages by industry and occupation which can be produced by the same piece of code. This implementation lives in the `PUMSMedianEconomics` class in `aggregate/PUMS/median_PUMS_economics.py`.

## Implementing on top of existing PUMS aggregation structure.

PUMS aggregation uses an inheritance tree where reused code to load data and iterate through indicators lives in the parent `PUMSAggregator` class. In general the best approach when implementing on top of this structure is to change code in parent classes to be flexible and handle different situations. Unfortunately with deadlines coming fast some janky fixes were added to the `PUMSAggregator` class to accommodate the new class. The goal is to straighten these out during refactor.

Additionally a new `PUMSMedian` class was added between `PUMSAggregator` and `PUMSMedianDemographics` but then was not used as a parent class of the new `PUMSMedianEconomics` class. Again these inconsistencies should be addressed during a refactor. 

## Double crosstab

Median wages here are crosstabbed occupation/industry and then also crosstabbed by race, making it the only indicator with a double crosstab. This is implemented with a nested for loop that first iterates through "economic crosstabs" (occupation, industry) and also through race. 

This is a rare situation in which indicators are used but not aggregated on directly. No counts, fractions, nor medians are called on occupation and industry directly. Previously all indicators assigned to the raw data were aggregated on, this pattern break is one among several was not handled robustly in this pull request and needs to be refactored. 

## Refactor assign economic indicators

The code to crosswalk the ACS variables to those laid out in the data matrix was moved to `assign_economic_indicators.py`. This refactor was good design it's called in multiple places. 

## Median Wages Notebook

Has my sanity checks

## Tests 

Simple straightforward tests were added to make sure wages are positive and PUMA-occupation combinations with no workers have NaN median wage as well

## Internal Review Files
Just the 5 testing PUMAs or "limited PUMAs". Full list of PUMAs takes an hour to run and I wanted to get this pull request in so that our fix to replicate weights can be incorporated. Full analysis of all PUMAs can run in background while I develop on other things


SashaWeinstein:
I want to flag this question that arose:
I read the "civilian employed population 16-64 **with earnings**; no percent calculated" in the denominator tab of the median wages in the data matrix as wages over 0. Maybe this was incorrect and this person should be counted and median wages for this data point will be zero. We should discuss

More context in the notebook

td928 (approved):
the new median workflow makes sense to me probably due to the similarities with the existing calculation for counts/fraction workflows. 

It is also helpful to see the new changes in economic indicator design since it will have an impact on the AMI work as well.

mbh329 (approved):
We need to be careful here with the `aggregate_PUMS.py` file as the branch `aggregate-PUMS-borough-citywide` also makes a change that adds in teh `geo_col` parameter in aggregation to the __init__. other than that the logic looks good. Haven't been able to run the new classes on my local machine due to the api issue

PR #135, opened 2022-02-15
Educational attainment indicator

SashaWeinstein:
Nothing novel here, one new function to assign the education field and a filter for those over 24


mbh329 (approved):
Straight forward changes, looks good

Issue #65, opened 2022-01-05
LI households


AmandaDoyle:
Description: Number of households 51-80% AMI

What year of AMI standards would you use for 2015-2019 ACS?
Do we have a chart of AMI by family size?
or earlier time point, would this be how many households qualified by earlier AMI standards or how many households qualify based on most recent standards? AMI bands based on most recent values or based on ACS 2019 dollar value?
Do not compare with 2000

Logic:

Source(s): ACS PUMS

Year(s):
- [ ] 2015-2019

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: Yes
Denominator: total households

SME: HPD/DCP Population Division

Questions:

Roadblocks:

Issue #55, opened 2021-12-29
ELI households


AmandaDoyle:
Description: Number of households that are <30% AMI

Logic:

Source(s): ACS PUMS

Year(s):
- [ ] 2015-2019

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: No
Denominator: total households

SME: HPD/DCP Population Division

Questions:

Roadblocks:

td928:
## 2022.2.4 Te's Work Journal

#### modules testing remain to do 
the new module testing cannot complete because PUMSEcnomics indicator weight calculation is broken. Need further investigation before concluding the new assign modules works

Follow the work done today [b388e26](https://github.com/NYCPlanning/db-equitable-development-tool/commit/b388e26885e544def087a24a6f44157074078678)

#### Solution (to dos)
this will involve create new ingest function (similar to one below) which after person level PUMS is ingested to transform the dataframe to keep only one member of the household as the representative. Since household income (HINCP) &  number of persons in family (NPF) are identical across different members. This should be fine.

https://github.com/NYCPlanning/db-equitable-development-tool/blob/b388e26885e544def087a24a6f44157074078678/ingest/load_data.py#L20-L27

then in the section below new logic needs to written to determine which ingestion functions to use depending on the indicators are on person level or household level. 

https://github.com/NYCPlanning/db-equitable-development-tool/blob/b388e26885e544def087a24a6f44157074078678/aggregate/PUMS/aggregate_PUMS.py#L55-L60

#### Alternative Solution (to dos)

create another table in the PUMSaggregator class to hold the household dataframe by same logic of keeping only the household members with "1" in . The place to add is probably in this section of the code 

https://github.com/NYCPlanning/db-equitable-development-tool/blob/b388e26885e544def087a24a6f44157074078678/aggregate/PUMS/aggregate_PUMS.py#L68-L75

but this will have downstream effects in part of the codes dealing with the . e.g. the add_counts currently use the self.PUMS as the default table to perform calculations on. With this approach, some of those functions will also need to be fixed. Might be more work.

https://github.com/NYCPlanning/db-equitable-development-tool/blob/b388e26885e544def087a24a6f44157074078678/aggregate/PUMS/aggregate_PUMS.py#L115-L130

PR #137, opened 2022-02-16
Households Income Bands Indicators and Households PUMS 

td928:
## Overview

As this branch started as work for only the ELI indicator, the scope has shifted considerably when all six indicators for number of household for ELI, VLI, LI, MI, MIDI, and HI which are issues #55, #64, #65, #66, #67, and #68. Since these are first set of indicators reported on the household level, while the additional lines codes are not that significant, but they are scattered across all portions of existing stacks. Thanks to the robustness and flexibility of the existing the stacks the new additions were able to slide in without much challenge. 

### Methodology Summary
Beside the creation of the PUMSCountHouseholds child class, which is largely modeled after the existing PUMCountEconomics class. Also, a household boolean flag  is passed inside the aggregator class and also in ingestion to ensure the proper loading . There are some designs probably should be reworked at the next refactoring. But it is currently not a priority.  The most important are listed below.

## PUMS ingestion
there are two major changes in the PUMS ingestion process are two. 
1. new variable_types created in the [variable generator](https://github.com/NYCPlanning/db-equitable-development-tool/blob/55-ELI-Households/ingest/PUMS/variable_generator.py) to pull only relevant variables for household income calculation and is opened to other addition for more household level indicators. 
2. new logics added to query manager to [pull the appropriate weights](https://github.com/NYCPlanning/db-equitable-development-tool/blob/c1c37b9998253efa31e8b4aed5547e5041c3ae88/ingest/PUMS/PUMS_query_manager.py#L90-L95) for the households level calculations

It should be noted once the queries are directed to pull households weights. The queries return only one member per household which saved us some work. 

## [PUMCountHouseholds](https://github.com/NYCPlanning/db-equitable-development-tool/blob/55-ELI-Households/aggregate/PUMS/count_PUMS_households.py)
The justification for creating this separate households aggregator class mainly because of the weight ingestion for household is separate from the personal weight ingestion. Therefore this aggregator's work do **not** mingle at all with other economics indicators such as industry or occupation calculation. 

## Test
some basic tests are involved including one pytest which passed but due to my inexperience with the pyest module cannot be very confident about their validity. This is the reason for the  I also created [a brief notebook](https://github.com/NYCPlanning/db-equitable-development-tool/blob/55-ELI-Households/notebooks/household_income_bands_checks.ipynb) for very similar tests. Testing is light on this work so far. So please let me know your suggestions on what other tests to perform.


SashaWeinstein:
Hey sorry I pulled down the branch to take a look at it in VSCode and fixed some merge conflicts Then I pushed the new commits without thinking about what it would do. Think it's ok but my apologies that was an oopsie

SashaWeinstein:
This looks great! I'm really impressed by how well you fit the new functionality into the existing structure. I got the code to work on my machine. I have some comments that I will break into three sections of ingest, aggregate and test

## Ingest

The new variable set for households is perfect. We can add a check so that the ingestion needs to be called with variable type as household if household bool is true and vice versa but that's just extra safety.

The filter for records where `SPORDER` is 1 is also great. I do think though that we can drop SPORDER in `assign_identifier` as it's called after `populate_dataframes`. Not necessary as the code works but might be nice to cut down on unnecessary columns. Same with `merge_rw`, I see why we need to keep `SERIALNO` but think we should drop `SPORDER`. Even better would be to make `SERIALNO` the index when household is true and ignore `SPORDER` completely once we've filtered on it.

I see that the replicate weights data types are changed in three places. In the `clean_data` method of `PUMSData` they are converted to integers which makes the most sense to me. In the `merge_rw` method `PUMSData` they are converted to strings which I don't understand and don't think should stay unless there is a good reason. Then again in `calculate_counts` they are sent to integers, a step is fine but redundant as it should be done during cleaning.

The variable_mapper was moved back to a constant in `PUMS_query_manager` where as it had lived in it's own file previously. It's defined twice in this pull request which I believe is due to poor git hygiene upstream. In my opinion should stay in it's own file to keep `PUMS_query_manager` more readable.

The way that `PUMS_query_manager` queries different weights based on what is passed in is quite readable. Think it would be better to have "weights" attr that is set to either "WGTP" or "PWGTP" on init but it's fine the way it is for now. 

To summarize, Te you seemed worried that ingestion the household level data would be a pain but I think you achieved it with relatively few lines of code so nice job. 

## Aggregation

First off want to comment that the `PUMSCountEconomics` class was pushed with a bunch of comments, **this file should definitely be reverted before merge**.

I also really don't like the unittest skip for the race fractions crosstabs in aggregate PUMS. If you're going to go with if/else blocks for the time being that's fine but this just needs one if that says 
```python
if not self.household: 
    for race in self.categories["race"]:
        etc
```
The in-line comments in `add_category` should be moved to the docstring as well

The new child class looks almost perfect, the only comment I have is that you don't need to copy the dataframe in the filter method, a copied dataframe is passed to the statistical functions so this code is redundant. 

Also the child class for household aggregator shouldn't allow the user to set the household parameter, it should just always initialize the parent class with `household=True`

## Tests 

I ran the code myself in ipython and it all looks good. Upper manhattan had highest fraction of extremely low income housing and greenpoint had the highest fraction of high income which is what I expected.

The tests makes sense to me. I would like to see at least one test on the aggregated level. Maybe check for a minimum number of total housing units per PUMA? Or check that no fraction is over 90% or 95% which we should treat as an impossibility if there are six categories.

td928:
Really appreciate your comment @SashaWeinstein! 

The `SPODER` is definitely a good catch. Actually if what I gathered about the `WGTP` endpoint is correct. The filter on it to be equal to 1 could even be dropped with no impact. But I am keeping it just in case. And I am restoring the drop in the `assign_identifier`. 

Another great catch for the type casting in `merge_rw`. This actually was intended as a fix when initially the first discovery some of the `SERIALNO` is mixed and would create issue in joining in the next step. However, this actually **CREATE** more problem down the line a.k.a the weight calculation errors I got! I agree the `clean_data` should take care the casting so I am moving it all there. Also I think it might be better to add the weights cleaner to `PUMS_cleaner.py` but I don't have a good enough understanding of it to do this. Would you mind helping out with this?

all the aggregator class comments make sense to me and I made all the necessary changes. I think when you said the `PUMSCountEconomics` comments you actually meant the `PUMSCountHouseholds`? I removed my comments in that file.

One aggregate test is added for checking whether the fraction indeed added up to 1.

td928:
> Hey I took a look and everything looks great. I one more question and a couple more comments.
> 
> My question is on: _"This actually was intended as a fix when initially the first discovery some of the SERIALNO is mixed and would create issue in joining in the next step. However, this actually CREATE more problem down the line a.k.a the weight calculation errors I got!"_ Not sure how the casting and SERIALNO relate? If it's important we can talk it through friday morning, if it's not important I'm happy not knowing.
> 
> _"Also I think it might be better to add the weights cleaner to PUMS_cleaner.py but I don't have a good enough understanding of it to do this. Would you mind helping out with this?"_ This is a great question. The `PUMSCleaner` does just one thing, which is apply custom logic to crosswalk a particular column. Casting one type to another doesn't really "fit" in the `PUMSCleaner` in my opinion. I think the solution is change the name of the `PUMSCleaner` to something more intuitive as to not get it mixed up with the `PUMSData` `clean_data` method. Should definitely keep the process as is and can worry about it during refactor
> 
> Here are a list of comments, all are short
> 
> * The .copy() in the filter of child class can come out
> * Inline comments in PUMSCount init call in child class can come out
> * don't need to append the "_household" string in `ingest/make_cache_filename`
> * remove the `local_variable_mapper` defined in the top of the PUMS query manager file, only need the one defined in it's own file

Hey good catch on all of those. Should be good now with the new commit. Thank you so much!

SashaWeinstein (approved):
All looks good to me! My feedback is in the conversation section

PR #143, opened 2022-02-18
Add citywide/borough columns to ingestion process, add geo_col to aggregation scripts,  and upload csv's for Citywide/Borough Level Geography's for ACS PUMS Data (08-12, 15-19) 

mbh329:
**Add Borough/Citywide to Ingestion process**

The base level geography that we initially aggregated the ACS PUMS data on was at the PUMA level (4 digit number) that specifies the unique county code (in our case - borough as these align with NYC counties) in the first two digits. We map these unique digits to a  borough level abbreviation in `utils/assign_PUMA.py` that follows the naming conventions as defined in this repo's wiki.

The code to add these two new geographies lives in the `ingestion/PUMS/PUMS_data` `PUMSData` base class in the nested `clean_data` function where we simply create two new columns in the data: `borough `and `citywide`. The `borough `column is created by applying the `puma_to_borough` function as mentioned above. We also set the `puma` column to reflect the lower case convention earlier in the data ingestion process. 

**Adding `geo_col` attribute to aggregation**

In order to add aggregation for all three geography's (PUMA, Borough, Citywide) we created a geo_col parameter in the `__init__` in the aggregator parent class (`PUMSAggregator`) and for the base aggregators: `PUMSCountDemographics, PUMSCountEconomics, PUMSMedianDemographics, PUMSMedianEconomics, PUMSMedians`.  This way you can just add a `'geo_col` parameter to any of class instantiation by passing a string with the level of geography ("puma" [default], "borough", "citywide")

**Internal Review CSV's**
This pull request also has CSV's for the Demographic and Household Economic Security indicators at the Citywide and Borough level for the full data universe for QA/QC purposes. For the external review, we will need to collate these files as mentioned in the indicators Epic.


SashaWeinstein (approved):
This all looks good to me

Issue #113, opened 2022-01-24
Educational outcomes: ELA


AmandaDoyle:
Description: percent of 3rd-8th grade students that are ELA  proficient by where students live

WWL Figure 5.75. Use WWL data for EDDT launch. DOE will supply annual data for future years.Cannot report last two years because of COVID.  Waiting on link/folder of that data from DOE

Logic:

Source(s): DOE; HPD 2017-2018 WWL data - spreadsheet [here](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/civic-engagement-studio/_layouts/15/Doc.aspx?sourcedoc=%7B3A897425-085E-4D69-9089-085A6366B421%7D&file=NTA_data_prepared_for_ArcMap_wCodebook.xlsx&action=default&mobileredirect=true)

Year(s):
- [ ] 2017/2018

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: Yes
Denominator: total students

SME: Chad

Questions:

Roadblocks:
Field name convention: `edu_ela`

Cwilliams-1:
Columns to be divided from 5_StudentPerformance tab in linked spreadsheet: H/B(All Students); I/C(Asian); J/D(Black); K/E(Hispanic); L/F(Other); M/G(White)


[https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/civic-engagement-studio/Shared%20Documents/03.%20Working%20Materials/03.%20%20Equity%20and%20Displacement%20Vulnerability/HPD-DCP%20EDDT%20Coordination/02.%20Indicators/EDDT%20QOL%20Indicators/DOE%20WWL%20Data/DOE%20Access%20to%20Opportunity%20Files/DOE%20Clean%20Data/Request%204%20and%205%20-%20Location%20of%20Proficient%20Schools%20and%20NTA%20Outcomes/NTA_data_prepared_for_ArcMap_wCodebook.xlsx?d=w3a897425085e4d699089085a6366b421&csf=1&web=1&e=JhPZXB]

td928:
@Cwilliams-1 Hey Chad is the right understanding that this is reported on NTA level instead of PUMA and we are not doing any translating for geographies?

td928:
@AmandaDoyle My bad. I now see that there are columns for just 4th grade and the other ones I assume are for 3rd to 8th grade.

Question for NTA vs PUMA remain

[...]

Issue #98, opened 2022-01-12
Educational outcomes: HS graduation rate


AmandaDoyle:
Description: HS Graduation rate by where students live (percent)

Logic:

Source(s): DOE, ACS - link to source data

Year(s):
- [ ]

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: 
Denominator:

SME:

Questions:

Roadblocks:

levysamu:
Source is **not** ACS. The ACS data (educational attainment) exists elsewhere in the EDDT and does not belong here.

We are confirming the DOE data (graduation rate defined as the percentage of students that entered 9th grade in 2014 and earned either a local or regents diploma in 2018) with them as we speak.

Outstanding questions:

- Can we get a more recent version?
- Can we display this over time?
- What geography?

Cwilliams-1:
Columns to be divided -- from 5_StudentPerformance tab in NTA_data_prepared_for_ArcMap_wCodebook spreadsheet in linked folder:
CT/CN(All Students);
CU/CO(Asian);
CV/CP(Black);
CW/CQ(Hispanic);
CX/CR(Other);
CY/CS(White)

https://nyco365.sharepoint.com/:f:/r/sites/NYCPLANNING/civic-engagement-studio/Shared%20Documents/03.%20Working%20Materials/03.%20%20Equity%20and%20Displacement%20Vulnerability/HPD-DCP%20EDDT%20Coordination/02.%20Indicators/EDDT%20QOL%20Indicators/DOE%20WWL%20Data/DOE%20Access%20to%20Opportunity%20Files/DOE%20Clean%20Data/Request%204%20and%205%20-%20Location%20of%20Proficient%20Schools%20and%20NTA%20Outcomes?csf=1&web=1&e=rV6lvK

levysamu:
Hey all, I'm looking at the output of this at the PUMA level and noticing a handful of PUMAs with a specific race/ethnicity high school graduation rate well over 100%. Thoughts on why this is happening and how we might want to handle it?

ANH>100 for 3903 and 4016
ONH>100 for 3810
WNH>100 for 4001 and 4106

Issue #87, opened 2022-01-12
Educational outcomes: math proficient 


AmandaDoyle:
Description: percent of 3rd-8th grade students that are math proficient by where students live

Logic:

Source(s): DOE; HPD 2017-2018 WWL data

WWL Figure 5.75. Use WWL data for EDDT launch. DOE will supply annual data for future years.Cannot report last two years because of COVID.  Waiting on link/folder

Year(s):
- [ ]

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: 
Denominator:

SME:

Questions:

Roadblocks: Waiting for source data

levysamu:
Years: 2018 (nothing newer because of COVID)
Race/Ethnicity: confirming with DOE
Denominator: We will get percent values directly from DOE and won't be reporting total values (confirming as well)

We also need to add ELA scores (this was a late addition to the data dictionary), but it uses the same data source from DOE.

Cwilliams-1:
Columns to be divided -- from 5_StudentPerformance tab in NTA_data_prepared_for_ArcMap_wCodebook spreadsheet in linked folder:
AR/AL(All Students);
AS/AM(Asian);
AT/AN(Black);
AU/AO(Hispanic);
AV/AP(Other);
AW/AQ(White)

https://nyco365.sharepoint.com/:f:/r/sites/NYCPLANNING/civic-engagement-studio/Shared%20Documents/03.%20Working%20Materials/03.%20%20Equity%20and%20Displacement%20Vulnerability/HPD-DCP%20EDDT%20Coordination/02.%20Indicators/EDDT%20QOL%20Indicators/DOE%20WWL%20Data/DOE%20Access%20to%20Opportunity%20Files/DOE%20Clean%20Data/Request%204%20and%205%20-%20Location%20of%20Proficient%20Schools%20and%20NTA%20Outcomes?csf=1&web=1&e=rV6lvK

PR #144, opened 2022-02-23
QoL Education Outcomes

td928:
## Overview
Three education outcomes indicators (#87 #98 #113 ) for quality of life are wrapped into this one pull request. They all reply on [the same source data](https://nyco365.sharepoint.com/:f:/r/sites/NYCPLANNING/civic-engagement-studio/Shared%20Documents/03.%20Working%20Materials/03.%20%20Equity%20and%20Displacement%20Vulnerability/HPD-DCP%20EDDT%20Coordination/02.%20Indicators/EDDT%20QOL%20Indicators/DOE%20WWL%20Data/DOE%20Access%20to%20Opportunity%20Files/DOE%20Clean%20Data/Request%204%20and%205%20-%20Location%20of%20Proficient%20Schools%20and%20NTA%20Outcomes?csf=1&web=1&e=rV6lvK) and logics are identical as well. 

## Methodology
all indicators output are in percentage where the either the number of proficient students or number of graduated students out of the total (tested) population or population of each race. 

### Calculate PUMA
Since the source data provides NTA level count and the population division crosswalk from NTA to PUMA is used for this calculation

### Borough & Citywide
the raw counts are aggregated first then percentage calculated for each categories. 

## Output
a quality of life folder under the internal review created and one education outcome csv is saved under each geography. I am open to change the naming of any of the indicators. I opt for the most verbose and descriptive and open to better ideas.


SashaWeinstein:
Taking a look now.

One thing I'm stumped on is that I have a `QOL/` folder in my machine that shows up in the `dev`  and `113-Education-Outcome-ELA` branches. It doesn't appear in github though. It's not in .gitignore. Any ideas and what's going on here would be appreciated

SashaWeinstein:
This looks pretty good, code is straightforward and easy to read. I think there are some improvements to be made to the .py file that I'm outlining below. Super curious to hear your thoughts 

### `get_education_outcome`
This is good, better to have indicator produced by function that can be called by the export process instead of via `__name__=="__main__"`. 
I think this function could be improved with some parameters. One is a boolean of whether to write to internal review. In the future we will want to call this for export process and not write to internal review.
The other is a parameter of what geography level to get. 

The variable called `df` that has the raw student performance data could be renamed to something more intuitive like `student_performance` or `raw_student_performance`. 

### `calculate_puma`, `calculate_borough`, `caluculate_city`
These functions are good, clean and readable. I think some improvement could be made pretty easily. 

The main improvement is that I think all of these can be collapsed into one function that takes the geography as a parameter and groups by that column. I think the dataframe of student performance should have it's "citywide", "borough", or "puma" column assigned in `get_education_outcome` and then the single calculate function just aggregates on whatever geography is passed to it. 

Other comments are minor
- Why call the groupby df `df_`?
- Don't need to copy `df_` when assigning to result
- Why create new column names with a convention that matches the source data? Why not something readable to us?
- What's with the to_list() on cols in assigning result in `calculate_city`? It looks to me like cols should be a list already

### `rename_fields`
This function I like a lot. Only two comments 
1. 2nd and 3rd `.rename()` calls should have inplace=True
2. Don't need to return None

Hope these comments are helpful, let me know if you want to talk through anything

td928:
Hey @SashaWeinstein 

Thanks! Some really great comments here and I incorporated a lot of them in this last commit but I have two questions. First is where do you see the use of `__name__=="__main__" because I don't see it on my end. Is it implied by the way it is written somehow? 

Second question is about the geo parameter for the main `get_education_outcome` function. I am not sure I see the value of giving it geo filter because I am not sure we will ever run this main function for only specific geography but not all three. 

also about using the column names from the spreedsheet I was hoping to leave the renaming to output columns to the last step. I was hoping the inline comment can shed some lights on their meanings but I don't think understanding the exact context of the column is crucial here anyway. Although if anyone does feel strongly about it being confusing, happy to change the order and refactor this.

td928:
@SashaWeinstein change made for adding the geography flag for the `get_edu_outcome` function

SashaWeinstein:
New changes look good! I would put this code 

```python
    df = df.merge(puma_cross[["NTACode", "PUMACode"]], how="left", on="NTACode")
    df.rename(columns={'PUMACode': 'puma'}, inplace=True)
    df['boro'] = df.NTACode.str[:2]
    df['citywide'] = 'citywide'
``` 
in `get_education_outcome` instead of `calculate_edu_outcome` but it doesn't matter much, that's just personal preference.

I also noticed that the column names end with `_pct`, our PUMS columns are `{indicator}-pct` (dash instead of underscore).

I also noticed that in PUMA 3710, graduation rate for asian non-hispanic is zero. This doesn't seem right? There are both NaN's and 0's in several graduation rates crosstabbed by race

AmandaDoyle:
@td928  Once we get final input from LS we'll update field names, but here's my feedback from what I've noticed
1) Update format of fields to agree with what's scoped out in the [field naming conventions](https://github.com/NYCPlanning/db-equitable-development-tool/wiki/New-data-pipeline:-field-naming-convention). Percent values should be in whole digits rounded to two decimal points
2) I spot checked a couple numbers everything looks good  despite some standing out and looking incorrect.
Example: for PUMA 4001 you have the as graduation_rate_2017_wnh_pct 1.10465, but that's the right calculation for the 4 NTAs BK72, BK73, BK76, BK90.

mbh329:
Also need to update puma values to have the leading 0 in outputs

td928:
Hi really appreciate everyone's feedback on this. the new commit should address them all with new naming conventions. Thanks! @SashaWeinstein @AmandaDoyle @mbh329

SashaWeinstein (approved):
This looks good! Couple minor things I would do differently but nothing worth holding up the merge. I appreciate you taking my feedback. 

Same issues with 0 graduation rate for some pumas when crosstabbed by race, but that's not issue with the code exactly

PR #146, opened 2022-02-23
Export aggregated PUMS

SashaWeinstein:
# Export Finished PUMS Indicators (demographics, household economic security) to Digital Ocean

## New workflow: `Export PUMS Aggregated`
This iterates through 3 geographies (puma, borough and citywide), 2 years (2012 and 2019) and 2 categories (demographics and household economic security). It calls one new .py file `external_review/external_review_PUMS` and a new case in the existing `export_DO.sh`  file. Everything else is re-used from existing workflows. 

It usually takes around 2 hours to run with all data. 

## External Review PUMS
This file is responsible for calling the new `load_aggregated_PUMS` and passing it variables from a `Typer` command line interface. It gets a dataframe of collated results for a single category and saves it to the `.staging` folder. 

## Load Aggregated
This was written to be similar to `load_data.py` in the ingestion folder. Similarly to how that file can deliver PUMS or HVS, this was written to deliver PUMS indicators and hopefully new functions can be written to deliver other ones as well. It has logic to look for a cached file and run the aggregations if no cached file is found. 

## Reorder and rename columns
All these changes are in `aggregate/PUMS/aggregate_PUMS.py`

Columns are reordered so that each category for each indicator has columns for it's counts and fractions grouped together, and then crosstabs on that category follow after each category is listed. Ages and income run from low to high, education runs from more educated to less educated. Other categories (industry, occupation) are listed alphabetically.

"Fraction" is changed to "pct" and variance abbreviations (cv, moe) are sent to lowercase. 

## Cache aggregated results
This upgrade doesn't effect the export directly but was necessary to test the external review code on my local machine. Aggregated data are saved to `.output` folder. There is no "requery" parameter like there is for the ingestion process, that's a possible upgrade. To re-run results just deleted cached .csv

Cache filenames are created in in `aggregate/aggregated_cache_fn.py`

While I was going through the cache process for raw PUMS to replicate it's structure I cleaned out some unnecessary parameters. 

## Remove Medians
All medians calculations we have are incorrect, they should be based on buckets. They are taken out of external review upload and exceptions are raised if the aggregation is attempted. 

### Miscellaneous Changes
Other workflows were modified to only be run when it's corresponding .yml is changed. This is because exports don't need to be run on push and we aren't getting helpful info back from test workflows

Tests were changed to conform to the new column name specifications. They weren't tested as some as out of date anyway

## Notes for future work

This utilizes two new gitignored folders, `data/` and `.staging/`. There may be simpler way to store these temporary files, possible place for refactor.

Additionally the existing external review and new load aggregated data processes both collating multiple indicators into a single category for export. This is poor design and should be refactored to be made consistent


SashaWeinstein:
Note that github action is failing with error code 137. https://github.com/NYCPlanning/db-equitable-development-tool/runs/5307165703?check_suite_focus=true

Google suggest it's a memory issue, trying with test data only to see if that works

SashaWeinstein:
This fails for economics, need to figure out a fix. Think it's a memory issue

PR #147, opened 2022-02-23
Access to Subway/SBS and Access to ADA Accessible Transit

SashaWeinstein:
These two indicators are similar so they are handled by the same file. They require minimal processing as they are already organized by PUMA by DCP transportation.


SashaWeinstein:
Two recent commits address what we talked about in standup. Folders renamed from "QOL" to "quality_of_life", and code to clean data moved from ingest to aggregate

SashaWeinstein:
Realized that pumas have leading zeros in internal review files, need to fix that before this gets merged

mbh329 (commented):
`access_transit.py `
logic looks good except for an issue with the pct calculations - the internal review files do not match the wiki conventions. They need to be multiplied by 100 (0.123 -> 12.3) 

`assign_PUMA.py`
 `puma_to_borough` fucntion has been updated to handle 5 character puma's (this function is widely used to aggregate from puma to borough).

SashaWeinstein:
Thanks for these comments Max, I will work on both these once the traffic indicator gets merged in as that work has code I can use for these issues

mbh329 (commented):
Just looping back to this pull req as I am going through the data dictionary. The internal review files should match naming conventions as outlined in the wiki and the issue:

-  lower case naming
- acesss-ADA-pct -> access_ada
- access-subway-pct -> access_subwaysbs)

SashaWeinstein:
Ah good catch I will work on those updates now

SashaWeinstein:
Ok @mbh329 addressed the column label issue in new commit

mbh329:
almost there, just need to go from ADA to ada in the column name

mbh329 (approved):

PR #148, opened 2022-02-23
GET request in while loop

SashaWeinstein:
Simple upgrade, based on screenshot Max shared over email. 

I tested it on my machine and it works. The logs show that some get requests take much longer to resolve, I think that's because they are bouncing back with an error code but it tries again. 

It's capped at trying five times, I capped it to limit long run times.


mbh329 (approved):
Great - glad this works!

PR #149, opened 2022-02-24
Traffic fatalities injuries

SashaWeinstein:
This pull request has logic to get average total injuries, pedestrian injuries, motorist injuries, cyclist injuries, and total fatalities over two 5-year periods. 
The ingestion of the datasets will ultimately be done through data library, that is a separate pull request.

Some code from the transit access to clean the source data was re-used and put in `utils/assign_PUMA.py`. That pull request has some issues with it's output that I need to address. One is whether PUMAs are supposed to have leading zero's. This output does include the leading zero. I plan to wait until this is merged in and then use the code in this `utils/assign_PUMA.py` to clean the transit datasets. 

Output makes sense, denser PUMAs have more injuries and fatalities. This data excludes highways and parkways which explains low numbers in outer boroughs where I assume a larger % of miles traveled by car are on highways and parkways


SashaWeinstein:
Pointed to main instead of dev

PR #153, opened 2022-02-25
Pedestrian hospitalizations

SashaWeinstein:
# Pedestrian Hospitalizations per 100k Residents
Pretty straightforward. Like other QOL indicators this is already aggregated, we just need to plug in numbers from csv we get from a link.

The metric we need is calculated for citywide and by borough, but it's calculated by community district instead of puma. The crosswalk xlsx file at [https://www1.nyc.gov/assets/planning/download/office/data-maps/nyc-population/census2010/nyc2010census_tabulation_equiv.xlsx] that Te used in the QOL educational for NTA to PUMA can also be used for CD to puma. 

I loop through a dataframe to get the dictionary from CD to puma. I try not use for loops to iterate through dataframes but it was the best way to move forward here. 

For PUMAs that span two CD's, I calculate an average between the two CD's and report that number.

Some CD's are null in the source data, the portal says this is for "confidentiality reasons".


SashaWeinstein:
Next I can work on getting the .csv in through data library if that's how we want to manage source data

td928 (approved):

td928 (commented):
I approved all earlier. But rereading Sasha's post in issue, I realize the puma is calculated by averaging a few coded from CDs. I am guessing the experts will need to weight in on this. @AmandaDoyle

td928 (commented):
also a few typos in the main script where the `internal` is misspelled as `interal`

mbh329 (commented):
This looks really good. functions in pedestrian hospitalizations work as expected on my machine and the outputs look accurate as well - I am assuming there 3 puma's that aren't reported in the original dataset?

Really like breaking up geography helper functions into more spec. scripts and functions. in a refactor, we should do a better job at commenting what each is for but they are pretty straight forward (I imagine we can use these in a lot of different places/different projects)

Is the `generate_2010_total_pop_table` getting population data from the decennial census but via a column in one of non-fatal assault spreadsheet we received?

SashaWeinstein:
The answer to your last question is yes. It's not the best way to get that info but it was the easiest. 

Glad you think the code is clear

PR #156, opened 2022-02-28
Add Script and internal review csv for Heat Vulnerability Index at the puma level

mbh329:
`Added heat_vulernability.py` script that loads and cleans the xlsx data (removes subboro area, clean colunm names and sets dtypes, adds leading 0 to puma col). The script also exports to a csv using the internal review function


mbh329:
@SashaWeinstein correct, can only aggregate at puma level for this indicator

SashaWeinstein (approved):
Looks good, approved

PR #158, opened 2022-02-28
Traffic fatalities injuries with correct denominator

SashaWeinstein:
# Use new street miles column to do robust calculation
This is the second pull request for this indicator after we realized we were doing it incorrectly.
We now use data emailed to us by dcp transportation that has street miles for each puma

## Back-solve for total number of events
The process to get all data for a given year range was decomposed into a new function, `get_year_range_df`. The code was also updated to return the total number of events, not the number per 100 street miles. Even though the number per 100 street miles is how the data comes originally and what we want as final output, a consistent process that does the division correctly requires the number of events and street miles.

## Calculate with total number and street miles
This math lives in mean_by_geography. The first parameter passed is a numerator dataframe of a single data point (pedestrian injury, cyclist injury, etc) across a given year range. The second parameter is the denominator dataframe of corresponding street miles for each year. Each is aggregated on the geography, and then the sum of the numerator divided by the sum of the denominator times 100 gives the number of events per 100 street miles.

### Check against previous results
I kept the internal review files from the previous process with the `_old` in the filename. You can see the numbers per PUMA are unchanged but the borough, citywide change a fair amount


mbh329:
Quick q: What's the reasoning to stage the traffic fatalities data in data-library but all the other data we are keeping in the resources folder? Does this have to do with privacy concerns around the data?

SashaWeinstein:
That is a great question I do not have a good answer to. Don't believe it's about privacy concerns. Was making this decision on a case-by-case basis but @AmandaDoyle has a better idea of the big picture.

In general it's ok to merge in work that points to the local resources folder and then make the data library upgrade later, I know that for sure

mbh329 (approved):
`traffic_fatalities.py`
Logic looks good - I was able to run the code recreate the csv outputs on my local machine and they matched the new csv outputs in the internal review file. column naming conventions and variable calls were clear throughout, concise code, and i think the refactoring for getting the year from its own function is solid.

`Internal Review` CSVs
Helpful to see the old and new outputs - good thing we cleared up the issues and yujin was able to give us the necessary columns for the geo aggregation. The safety_1620_trafficfatalities seem quite low across the board but I assume this might be a product of not counting the highways, etc. in the final numbers

mbh329:
Side note: I think @AmandaDoyle wants us to connect the pull req to its respective issue to keep better track of the project as we complete issues

PR #159, opened 2022-02-28
finished covid death indicator

td928:
# Covid-19 Death Indicator

## Overview
apology for the extra large commit for this indicator and will try to do a better job showing steps in the future. It is overall fairly simple indicator. All data come from single excel sheet where total covid death is provided for each ethnicity on puma level. 

## Methodology
[Deaths by race and PUMA_20220202.xlsx](https://github.com/NYCPlanning/db-equitable-development-tool/files/8092165/Deaths.by.race.and.PUMA_20220202.xlsx) is read in and first cleaned before transformation by the `covid_death` function. 

The overall flow is very much modeled after Sasha's work on pedestrian death which I think makes a lot of sense for most quality of life indicators and I would probably refactor my other works e.g. education outcome like this in the future if possible. 

### covid_death
Once the cleaned dataframe is loaded. The logic either sends it to the citywide/borough where a groupby geography is performed before the pivoting transformation or direct pivot in the puma. 
`final[''] = final.sum(axis=1)` is needed to create the total without race breakdown before the final reformatting and write to internal review step which also takes advantage of the new functionality from pedestrian death work (not merged in yet). 

### load_clean_source_data
only caveat in this part of the code is the source data contains a few censored data point where the death fell below 5 for the puma. For the final output, I swapped out those `*` with `np.nan` for calculations. 

## Review
some spot checks for total were performed for each level of geographies. Maybe some additional testing could be done on PUMA to look at it spatially. Please let me know if it would be helpful to check in a notebook. Thank you!


SashaWeinstein (commented):
This all looks quite good Te. The one substantial comment I have is to take away the if/else block that skips the groupby if the geography is puma. Grouping on the puma does nothing, it returns a dataframe with the same information. But the code further down still works and my preference is to do the unnecessary groupby to reduce complexity. That's the way my indicators are written.

Other than that I have only one piece of style feedback: change the string concatenation in `columns={col: indicator_col_label + col}` to string formatting, something like `columns={col: f"{indicator_col_label}{col}"}`. This is more readable imo.

There are is one commented out print statement to clean up too. 

I like adding the blank column with the sum and then giving it the correct column name with the for loop, neat trick

mbh329:
I think we need to be very careful with this indicator and this should be double checked by subject matter experts and higher ups at DCP. From a quick google search of covid deaths in NYC from google I saw that we are underreporting the data by almost 2000 deaths (the data on google is updated daily so we are bound to have of discrepancy as our data comes from early Feb) but a difference of 2000 deaths seems a lot.

td928:
Hey @SashaWeinstein  thanks for the comments. Please take a look at the new changes when you have a chance and let me know if they all look good. Thanks

td928:
Hey @SashaWeinstein  thanks for the comments. Please take a look at the new changes when you have a chance and let me know if they all look good. PUMA helpers also changed.  Thanks.

SashaWeinstein:
This all looks good, but why was the `community_district_to_PUMA` function deleted from `PUMA_helpers.py`? I need that for pedestrian hospitalizations I think

td928:
> All my suggestions have been incorporated so that looks good. But why was the `community_district_to_PUMA` function deleted from `PUMA_helpers.py`? I need that for pedestrian hospitalizations I think

I think this is due to my version of assign_PUMA came from the dev which is as step or two behind. Would this get automatically sorted out when the other pull requests are merged in? Or should I add manually in here?

SashaWeinstein:
Interesting, ok. Yea I think that's makes sense and it will come back with it's associated pull request. Not good git hygiene though somewhere along the line, gives me small anxiety

PR #160, opened 2022-02-28
Access employment

SashaWeinstein:
Didn't route through data library, can as a next step

Also the numbers don't match the table in the issue, but they do seem to be correct based on the numbers in the .csv


aggregate/quality_of_life/access_to_jobs.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def access_to_jobs(geography, write_to_internal_review=False):
+    indicator_col_name = "access_employment"
+    clean_df = load_clean_source_data(indicator_col_name)
+
+    final = clean_df.groupby(geography).sum()[indicator_col_name]
+    if write_to_internal_review:
+        set_internal_review_files(
+            [(final, f"access_employment.csv", geography)],
```
td928, line 16: not sure the format string is necessary here but this is really minor.

td928 (commented):
Hey @SashaWeinstein really like this short and sweet. My comments should be brief as well. 

First one I made as comment inline. The second one is about the use of assign_PUMA. Should it be updated to PUMA_helpers before merging with dev?

aggregate/quality_of_life/access_to_jobs.py
SashaWeinstein: oh yea lol you're correct I can fix that. pure habit I guess?

SashaWeinstein:
I think the easiest is to merge in work to dev that renames the file and then merge that back into this branch? Easier than changing stuff myself

td928:
@SashaWeinstein my pull request for covid death is now merged in. Let me know if this gave you the path to move forward with merging the changes in for this. Thanks!

SashaWeinstein:
Ok @td928 take a look when you have a chance, merged in the new work and the puma helpers live in the .py file with the correct filename

td928:
Hey @SashaWeinstein all look good on access employment work. I did quick general sweep and found the access to open space also needs to switch to PUMA_helpers.  I don't think it needs its own issue so I did it and made push. Let me know if it looks good to you as well and I think it is ready to merge

td928 (approved):

SashaWeinstein:
Awesome, I don't see anywhere else where the old assign_PUMA was referenced when I used the search feature so we should be good to go

PR #161, opened 2022-02-28
Export PUMS Indicators to Digital Ocean

SashaWeinstein:
# Export Finished PUMS Indicators (demographics, household economic security) to Digital Ocean

## New workflow: `Export PUMS Aggregated`
This iterates through 3 geographies (puma, borough and citywide), 2 years (2012 and 2019) and 2 categories (demographics and household economic security). It calls one new .py file `external_review/external_review_PUMS` and a new case in the existing `export_DO.sh`  file. Everything else is re-used from existing workflows. 

This workflow is running into issues with memory, the limited puma work but the full dataset has not. 

## External Review PUMS
This file is responsible for calling the new `load_aggregated_PUMS` and passing it variables from a `Typer` command line interface. It gets a dataframe of collated results for a single category and saves it to the `.staging` folder. 

## Load Aggregated
This was written to be similar to `load_data.py` in the ingestion folder. Similarly to how that file can deliver PUMS or HVS, this was written to deliver PUMS indicators and hopefully new functions can be written to deliver other ones as well. It has logic to look for a cached file and run the aggregations if no cached file is found. 

## Reorder and rename columns
All these changes are in `aggregate/PUMS/aggregate_PUMS.py`

Columns are reordered so that each category for each indicator has columns for it's counts and fractions grouped together, and then crosstabs on that category follow after each category is listed. Ages and income run from low to high, education runs from more educated to less educated. Other categories (industry, occupation) are listed alphabetically.

"Fraction" is changed to "pct" and variance abbreviations (cv, moe) are sent to lowercase. 

## Cache aggregated results
This upgrade doesn't effect the export directly but was necessary to test the external review code on my local machine. Aggregated data are saved to `.output` folder. There is no "requery" parameter like there is for the ingestion process, that's a possible upgrade. To re-run results just deleted cached .csv

Cache filenames are created in in `aggregate/aggregated_cache_fn.py`

While I was going through the cache process for raw PUMS to replicate it's structure I cleaned out some unnecessary parameters. 

## Remove Medians
Some medians calculations we have are incorrect, they should be based on buckets. They are taken out of external review upload and exceptions are raised if the aggregation is attempted. 

## GET Request in while loop
This change was implemented to get past those periods where the API throws errors for no discernible reason.

### Miscellaneous Changes
Other workflows were modified to only be run when it's corresponding .yml is changed. This is because exports don't need to be run on push and we aren't getting helpful info back from test workflows

Tests were changed to conform to the new column name specifications. They weren't tested as some as out of date anyway

## Notes for future work

This utilizes two new gitignored folders, `data/` and `.staging/`. There may be simpler way to store these temporary files, possible place for refactor.

Additionally the existing external review and new load aggregated data processes both collating multiple indicators into a single category for export. This is poor design and should be refactored to be made consistent


mbh329 (commented):
Lot of stuff going on here. A lot of the changes to the 34 files are minor - either cleaning up parameters (i.e. removing household parameter/or defaulting household=False), linting the code to style and match bandit conventions, changing statistical column name conventions from upper case to lower case, change fraction to pct as outlined in wiki.

Overall the logic looks good and I like that it follows similar logic in older code. I think the `aggregated_cache_fn.py` and the `load_aggregated.py` are good additions but don't have much of an opinion on where the` load_aggregated.py` file should live - makes sense to me that it lives there for now. 

Clear comments throughout which was helpful (i.e. the different `order_columns` code and how they differ)

I went through all the files and wanted to post some questions:

- Does poetry automatically update the version of python? In the toml file, it specs the version of python differently and was curious if you have to explicitly set that because of some dependency issues. 
- Can you remind me what denominators we use for the education indicators (is that population over 25)? We institute that change by filtering the pop for those over 25 correct?
- Why is the `indicators_denom` tuple in the `count_PUMS_ecconomics.py` changed? Does that have to do with reordering the columns?

SashaWeinstein:
These are great questions

 - I wish I remembered why I had to specify the python version. Seems to have something to do with scipy 
 - Correct! The `indicators_denom` attr is populated with the tuple of `("education", "age_over_24_filter")` which points the parent class to the `age_over_24_filter` method 
 - I believe I reordered the indicator_denom lists to match the order from the data matrix

td928 (commented):
I think it is really impressive to see this work starts to flow top to bottom. Nice work @SashaWeinstein to pull it all together. 

Some comments I want to share which I will continue to edit for a little bit.

### Naming Convention
I am still unsure about the naming convention. In this work race, moe, cv suffix is appended with `-` and in the [wiki](https://github.com/NYCPlanning/db-equitable-development-tool/wiki/New-data-pipeline:-field-naming-convention) it seems that all should be appended with `_`.

mbh329:
Thanks for answering all those questions Sasha!

mbh329 (approved):

SashaWeinstein:
Ah thanks for this comment Te, sorry I missed it. I can address this at beginning of next week

PR #164, opened 2022-03-01
Test QOL

SashaWeinstein:
# New Tests
I had some time yesterday when I had finished the access to jobs indicator but was too tired to start a new one. So I wrote some tests that call the indicator functions we've merged in so far and make sure that the data has the correct indices. 

These tests all passed easily because the indicators had been written in a consistent way 🎉  (good job everyone.)

I had to make some minor changes to existing indicators to get them to work but I didn't change any of the logic.

# Next step for export
The next step for exporting is adding to the `accessors` dictionary in `external_review/external_review_collate.py`. This is how we loop through the indicator functions and get outputs to send to digital ocean. 

## .staging folder
The export PUMS process that I opened a pull request for implements a process where .csvs are written to a .staging folder and then copied to digital ocean. I think this is good design, take a look at that pull request and we can discuss how we want to export


mbh329 (approved):
Thanks for making those changes to everyones script and updating the clean_PUMAs function. Everything looks good to me

Issue #96, opened 2022-01-12
Health outcomes: Overdose


AmandaDoyle:
Description: Overdose Rate per 100,000

Logic:

Source(s): Need link to source data

Year(s):
- [ ]

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: 
Denominator:

SME:

Questions:

Roadblocks:
- Epiquery only shows city-wide. Need to follow up with DOH

Issue #95, opened 2022-01-12
Health outcomes: Diabetes


AmandaDoyle:
Description: Percent of adults; 18-64 with diabetes

Logic:

Source(s): https://www1.nyc.gov/site/doh/data/data-sets/community-health-survey-public-use-data.page

Year(s):
- [ ]

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: 
Denominator:

SME:

Questions:
- What years are we reporting data for?
Roadblocks:
- Might need help getting data in a format that we can work with.

td928:
similar to self-reported health. The confidence interval question and also whether UHF by race breakdown is available.

Issue #61, opened 2021-12-29
Health outcomes: infant mortality per 1,000


AmandaDoyle:
Description: infant mortality per 1,000

Logic:

Source(s): https://www1.nyc.gov/site/doh/data/data-sets/vital-statistics-data.page

Year(s):
- [ ]

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: No
Denominator:

SME: DOHMH

Questions:
- Do we need a special cut for the data from DOHMH to get data at the PUMA level?
- 
Roadblocks:

SashaWeinstein:
There are no years specified in the issue but the source data appears to have 3 separate year ranges. It has 2000-04, 2010-14, and 2015-19

PR #169, opened 2022-03-03
61 infant mortality

td928:
## Overview

Three indicators for mortality (#61 #95 #96) combining into one pull request. 

The three indicators work functionally identical and share same two source files. Some caveats for the script and also the indicators I will lay out below. 

## load_clean_source_data
This really took inspiration from Sasha's work on safety and read the source files by passing an arguments dictionary. 

## rename_columns
Almost the entirety of this work takes place in this little function where indicator name, race naming convention, and year range (for puma) are all remapped. Then finally the sort of clumsy order adjustment to conform to our naming convention. I really wish I had a more elegant solution than the monstrous list comprehension but the tricky thing is to account for both the column with race breakdown and total. Open to suggestion on improving this. 

Another caveat to point out here is the puma level data is provided in year range (2000-2004 etc. ) where the borough and citywide data seems to be on specific year (2000, 2010, 2019)

## null data
Many censored data point throughout the spreadsheets which are being replaced with `np.nan`. This is the same convention I followed as other quality of life indicators.


SashaWeinstein:
This looks pretty good. 

I like the race mapper iteration, I think that's the right approach.

I think time periods go after indicator names, so health_2000_infantmortality_wnh should go to health_infantmortality_wnh. Also the non-race crosstabbed indicator goes before it's race crosstabs. 

There is this weird `~$QOL_health_infant_premature_overdose_borough.xlsx` file in the QOL resources, probably generated in error and can be deleted

I didn't realize that the years don't match across the geographies, we have start year for 2 of the 3 year ranges but end year for the last (2019). 

Finally I have an idea of how to make the column reorder more readable. If we need to produce a list, we could start with an empty list and write for loops to populate it? Can pair code this if it's helpful

AmandaDoyle (commented):
Confirmed: Borough and Citywide reports data at 2000, 2010, 2019
PUMA-level data rolled up 2000-2004, 2010-2014, and 2015-2019 to enable getting rates for race/ethnicity + all at the PUMA-level.

AmandaDoyle (approved):
The outputs in terms of column names and data look good to me.

PR #170, opened 2022-03-03
Covid deaths as rate by 100k

SashaWeinstein:
Built on work that Te did. Get 2020 population for each race from Erica's numbers. Then get the rate per race. Finally then sum up the cases per race and population by race to calculate rate for all people in a geography.


AmandaDoyle (approved):

td928 (approved):
Hey @SashaWeinstein definitely a more complicated work than it initially looked. 

I think it is irrelevant for approving this particular work but I think it would be a good flag for the future to add calculate 100k rate to statistics as a function/module. Similarly for the census 2020 population work. I am not entirely sure if it would come up again but still a valuable module. 

Overall all look good to go. Thank you!

SashaWeinstein:
Yep you're 100% right, that's an important refactor. Wrote [an issue](https://github.com/NYCPlanning/db-equitable-development-tool/issues/172) and assigned us both

PR #171, opened 2022-03-04
151 rename housing production attributes

td928:
## Overview
All output names updated to follow new conventions for both the change in units indicator and also the affordable units from HPD indicator. 

## Refactoring change_in_units
Three functions in change in units reduced to one for better design and reduce about 40 lines of codes. 

## HPD affordable script
The renaming for columns take place in the final transformation. A simple approach of replacing the appropriate strings was taken. 

## Output files 
all output saved to internal review


aggregate/housing_production/hpd_housing_ny_affordable_housing.py
```diff
@@ -59,6 +59,14 @@ def load_housing_ny():
 
 
 def pivot_and_flatten_index(df, geography):
+    level_mapper = {
```
AmandaDoyle, line 62: @td928  Question.  To change `eli_units` to `units_eli` would you just need to adjust this snippet of code?  I was thinking about this last night and kicking myself because I think reversing units and the income band would better follow our convention.  If you'd hate me for making this change then don't worry about it.

td928: Not a problem at all! This should be an easy change

AmandaDoyle (commented):
Reviewed data values and those look good to me.

SashaWeinstein (approved):
This all looks good. The refactored functions should still work with the export process, and if output data is the same we can be confident that it didn't mess anything up.

The new code to change column names looks good, I like declaring dictionaries in the module namespace and referencing them in functions. Having a consistent way to declare the final indicator column labels makes the the columns easier to change down the road, which was clear right away when Amanda asked for that change

td928:
Hey @AmandaDoyle @SashaWeinstein Thank you both for your comments. I believe the total units 2010 has been addressed. Also the new commit swapped the position now it is `units_eli` etc.Please checkout the new output files when you have a chance!

SashaWeinstein:
This looks just to be code to rename an existing column, does it also add it to the final output?

AmandaDoyle (approved):
Approved with new field names

Issue #80, opened 2022-01-10
Demographic Indicators


SashaWeinstein:

AmandaDoyle:
AD feedback on output 

- [ ] 1) Update file names and folder structure to match what was scoped out in the epic so that structure is consistent across outputs.
- [ ] 2) Where is the 2000 data?
- [ ] 3) Something is wrong with the outputs because there are only 6 rows in the demographic 2019_by_puma.csv
- [ ] 4) Remove `total_pop-pct-denom` field.
- [ ] 5) I think the the denominator logic is different for the race and hispanic origin category.  Instead of having the denominator be the total population of that race, it's the total population (i.e. pop anh pct would be 4677 / 110196 = 4.24%)
- [ ] 6) Remove the `total_pop-anh-pct-denom` and other race denominator fields for total pop.  Let's discuss how to best record race
- [ ] 7) Double check to make sure that CV calculation matches was was done in Pop Fact Finder.  See [here](https://github.com/NYCPlanning/db-factfinder/wiki/Estimates,-MOEs,-and-CVs#calculating-coefficient-of-variation)
- [ ] 8) [This](https://github.com/NYCPlanning/db-factfinder/wiki/Median-calculations) document walks you through of how to calculate median age
- [ ] 9) I checked the counts, MOEs, and percent values for 2019 data for PUMA 3701 using the census summary tables and the values look correct to me.
- [ ] 10) We'll need to rename the fields, but lets draft out the naming convention and agree on it so were on the same page before implementing anything

SashaWeinstein:
Hey sorry about the 6 columns in the by puma, I thought I ran the export with all data but I only ran with test data. I will re-run now and then we will have the full columns. Will work through other feedback while that is running

SashaWeinstein:
2000 data is from Erica, we are holding off on that

SashaWeinstein:
Putting Te's comment here 

# Naming Convention
I am still unsure about the naming convention. In this work race, moe, cv suffix is appended with - and in the [wiki](https://github.com/NYCPlanning/db-equitable-development-tool/wiki/New-data-pipeline:-field-naming-convention) it seems that all should be appended with `_`.
# 
[...]

PR #175, opened 2022-03-07
Take off total pop race crosstabs

SashaWeinstein:
# Fix race crosstab of total population
This fix is in response to the fifth checkbox in [Amanda's long comment on the demographic indicator](https://github.com/NYCPlanning/db-equitable-development-tool/issues/80#issuecomment-1050981122). She correctly noticed that the `anh` percent was 100, when it should have been much lower. 

This information on percentage of total pop by race won't ultimately come from PUMS, it will come from the decennial census. We do use the info but only to calculate percentages, and it's reported in the denominator column. This means there is no good reason total pop by race should have it's own indicators with associated variances and percentages. 

## Changes to code
The changes are rather janky, just adding if statements to not iterate through races for total pop and then not including total pop in the logic to order the columns at the end. A better approach is to set total pop as a different sort of indicator that doesn't get crosstabbed by race. This make the code cleaner. But to hit deadlines a faster approach was taken 

### Question
Can we get total pop from decennial census? If so total pop can be removed as indicator here and a small but persistent headache goes away


aggregate/PUMS/aggregate_PUMS.py
```diff
@@ -236,10 +237,22 @@ def add_category(self, indicator):
...
+            if ind_denom[0] != "total_pop"
+        ]
+        # Don't love hardcoding the beginning of this list, can be refactored
+        col_order = [
+            "total_pop-count",
+            "total_pop-count-cv",
+            "total_pop-count-moe",
+            "total_pop-pct",
+            "total_pop-pct-cv",
+            "total_pop-pct-denom",
```
td928, line 252: quick reference here to [one of Amanda's comment](https://github.com/NYCPlanning/db-equitable-development-tool/issues/80#issuecomment-1050981122) about leaving out the `total_pop-pct-denom` @SashaWeinstein

td928, line 252: quick reference here to [one of Amanda's comment](https://github.com/NYCPlanning/db-equitable-development-tool/issues/80#issuecomment-1050981122) about leaving out the `total_pop-pct-denom` @SashaWeinstein

SashaWeinstein: yep, you're right about that. good catch

SashaWeinstein:
I think per our conversation at monday standup this can be further improved to not include total pop columns at all? Those come from the decennial census. I can work on that this morning, should be quick

SashaWeinstein:
Ok new changes take away total pop as indicator, it won't be reported at all in final output. 

The only downside of the new code is that the total pop was good to double check denominators. We could make adding it optional for debug but maybe that's overkill

td928 (approved):

PR #179, opened 2022-03-08
demographics cv calculation

td928:
## Overview

very simple work on the CV to follow the recipe in PFF. Apology for the very bad numbering for this branch which was directed at the issue #80 . I also created a separate issue specific to this work #179 .

## variance_measurement
added to existing MOE calculation file as a function, the cv calculation is run with the variance_measurement function. 

## calculate_count and calculate_fractions
Where the `variance_measurement` is run are moved more upstream immediately after the survey package is run. This design helps to simplify the later pivoting and renaming operations later. Also it is noted that moe was not properly added to the final results 

## keep_SE
The keep_SE flag is used to decide whether to drop all `se` columns at the end of calculation. However, it should be flagged here that the `se` columns are not included in the current implementation in `dev` regardless of the value of this flag because the final reorder columns do not include `se`. See code below for reference. It is a bigger design decision on whether the flag should be incorporated in the PUMAggregator class to keep those columns. 

https://github.com/NYCPlanning/db-equitable-development-tool/blob/1adfe2ddd1674816c5a7c4728eeecfe697ef96f2/aggregate/PUMS/aggregate_PUMS.py#L243-L249

### DIFF
--- a/statistical/MOE.py
+++ b/statistical/MOE.py
@@ -6,26 +6,22 @@ import pandas as pd
 from scipy import stats
 import numpy as np
 
-z_score = stats.norm.ppf(0.9)
-
+z_score = stats.norm.ppf(0.95)
 
 def variance_measures(df, add_MOE, keep_SE):
     if add_MOE:
         df = SE_to_MOE(df)
-    if not keep_SE:
-        df = remove_SE(df)
+    df = add_CV(df) 
     return df
 
 
 def SE_to_MOE(df):
-    for c in df.columns:
-        if "-se" == c[-3:]:
-            var = c.rsplit("-", 1)[0]
-            df[f"{var}-moe"] = df[c] * z_score
-
+    df['moe'] = df['se'] * z_score
     return df
 
+def add_CV(df: pd.DataFrame):    
 
-def remove_SE(df: pd.DataFrame):
-    df.drop(columns=[c for c in df.columns if c[-3:] == "-se"], inplace=True)
-    return df
+    df['cv'] = df['se'] / df['V1'] * 100
+
+    return df 
+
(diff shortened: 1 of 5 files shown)


statistical/calculate_counts.py
```diff
@@ -59,21 +60,26 @@ def calculate_counts(
         by=data[[geo_col, variable_col]],
         design=survey_design,
         FUN=survey_package.svytotal,
-        vartype=base.c("se", "ci", "var", "cv"),
+        vartype=base.c("se", "ci", "var"),
     )
+    #print(aggregated)
+    aggregated = variance_measures(aggregated, add_MOE)
```
SashaWeinstein, line 66: this is a great upgrade, makes the code to assign variance columns much simpler

statistical/MOE.py
```diff
@@ -6,26 +6,22 @@
...
     return df
 
 
 def SE_to_MOE(df):
-    for c in df.columns:
-        if "-se" == c[-3:]:
-            var = c.rsplit("-", 1)[0]
-            df[f"{var}-moe"] = df[c] * z_score
-
+    df['moe'] = df['se'] * z_score
```
SashaWeinstein, line 19: so much easier!

SashaWeinstein:
You're totally right ab the keep SE flag not doing anything. we should remove that and the add MOE flag. Added an enhancement in [this issue](https://github.com/NYCPlanning/db-equitable-development-tool/issues/180)

SashaWeinstein:
I ran the code on my machine and all looks good. made one change which is to send the z-score back to 90% confidence interval. I also changed the name of a file.

SashaWeinstein:
Also Te your linter doesn't seem to be set up right because the code you're pushing still uses single instead of double quotes.

mbh329 (commented):
The logic looks good to me, I am just testing this now on my local machine

mbh329 (approved):

Issue #182, opened 2022-03-08
Order columns out of `PUMSAggregator`


SashaWeinstein:
This method should be moved to `PUMSCount` as it's specific to counts/percentages. Not sure how to only call the function for that class but it can't be too hard

PR #184, opened 2022-03-08
176 linear interpolation medians

SashaWeinstein:
# Calculating medians the correct way
All work in this pull request is based off population factfinder's median calculations which can you can read about [here](https://github.com/NYCPlanning/db-factfinder/wiki/Estimates,-MOEs,-and-CVs#calculating-coefficient-of-variation). I tried to follow that code as closely as I could, but there are some changes. The biggest change is that I work off a dataframe that is arranged differently than the row used in pff. Additionally throughout the code there are small differences in syntax that I introduced. I wanted to make comparing the functions easy but also use patterns I'm familiar with.

I don't elegantly handle any corner cases like the pff code does. Fortunately none of the pumas have the corner cases for the age variable. For household income and wages we may get more but we will cross that bridge when we come to it.

### Integrating new functions into aggregation pipeline
I only call the new function from demographic medians as 1) it's the simplest (best place to start) and 2) demographics is the category we are trying to get out tomorrow.

### Internal Review
There is one internal review file with only the 5 test PUMAs. Calculations using this method are much quicker than the survey package. I checked all 5 of the test PUMAs against the summary files and they fell within the margin of error.

I couldn't get all PUMAs because I locked myself out of the census API by making too many requests. There was an error with the while loop that is addressed in this PR. Hopefully tomorrow we can run the full dataset of pumas

### Metadata
I get the bins and design factor from a json which is how pff does it. I copied the json straight from that repo as the bins in the data matrix match.

### Miscellaneous changes
The order_columns passed to base aggregator init is jank. I wrote an [issue](https://github.com/NYCPlanning/db-equitable-development-tool/issues/182) for an enhancement that is better design.

Another change is that medians don't download PUMS with replicate weights. I didn't get a chance to test this code on my machine as the API isn't working for me today but hopefully it will work on someone's machine

### PUMS GET
Fixed this, counter ticks up to only make 5 requests per attempt.


statistical/calculate_medians_LI.py
```diff
@@ -0,0 +1,131 @@
+"""Calculate medians using linear interpolation"""
+import json
+from scipy import stats
+import numpy as np
+import pandas as pd
+
+z_score_90 = stats.norm.ppf(0.9)
```
td928, line 7: wonder why this was changed to 90 again. Since the only place I am seeing it being used is the cv calculation. I thought we settled the .95 is the correct one to use.

td928 (commented):
more a question than a comment. I look at the PFF repo and it does seem that they are using data without dealing with the replicate weight in general and we are doing this similarly for our median calculation. I wonder if there are any differences between these two streams of querying from PUMS?  My understanding of the PFF source data are based on [this script](https://github.com/NYCPlanning/db-factfinder/blob/292ffb038c09c54a4e15c298a9b7c6197e4c1965/factfinder/download.py#L14)

SashaWeinstein:
Yes Te that's correct, replicate weights aren't used in linear interpolation. The code should be changed so that the PUMS data used by the median PUMS demographics base aggregator doesn't have replicate weights

statistical/calculate_medians_LI.py
SashaWeinstein: ok went back to Erica's message and see that 1.645 is correct z-score to use, pushing this change now

td928 (approved):

PR #189, opened 2022-03-10
Decennial Census For Export into dev

mbh329:
Because there is a roadblock on the 2000 pums data, we are creating a new pull request just for the decennial census data so that it can be merged into and exported to DO.


SashaWeinstein:
awesome! making some changes to the design of the export process but keeping the logic implemented here

Issue #91, opened 2022-01-12
Health outcomes: Self Reported Health


AmandaDoyle:
Description: Percent in each category for Self Reported Health (percent poor, good, very good, excellent)

Logic:

Source(s): Need link to source data

Year(s):
- [ ]

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: 
Denominator:

SME:

Questions:

Roadblocks:

td928:
@AmandaDoyle just to have list of questions about this indicator for you. 

1. Is confidence interval is included in the final indicator? How should them be handled?
2. the NTA level report on the 'UHF' tab does not have a racial breakdown. Is it not available and won't include for puma level?

levysamu:
Hey all, I'm looking at the output of this at the PUMA level and noticing a handful of PUMAs with a specific race/ethnicity high school graduation rate well over 100%. Thoughts on why this is happening and how we might want to handle it?

ANH>100 for 3903 and 4016
ONH>100 for 3810
WNH>100 for 4001 and 4106

PR #191, opened 2022-03-10
Self Reported Health and Diabetes

td928:
## Overview

with the new information we have on those two final health indicators (#91  #95 ) we should be able to wrap things up.  Work is very simple now with the new clean spreadsheets.

## PUMA
The only wrinkle in this work is the casting from CD provided in the spreadsheet into pumas. Our `community_district_to_PUMA` works but it takes additional steps to format the CD column into the borough code + digit(s). Also since the information is actually provided on PUMA level just rewrapped as CDs, a drop duplicate is needed to get to the final puma output. I did spot checks for all duplicate puma and results seem fine. 

## Improvement
The two functions for each indicators are largely identical since they all come from identical spreadsheets. I can make arguments for both sides to either keep as it is for easy to access from the export function or reduce redundancy in this file.


internal_review/quality_of_life/self_report_health.csv
```diff
@@ -0,0 +1,4 @@
...,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2006,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2007,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2008,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2009,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2010,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2011,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2012,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2013,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2014,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2015,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2016,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017,2017
```
AmandaDoyle, line 1: What is this file?  Is this needed?

AmandaDoyle (approved):
Outputs look good so I'm approving; however, I have one question of if the self_report_health.csv is needed.

internal_review/quality_of_life/self_report_health.csv
td928: good catch! It was probably relic from the past work. Removed in the new commit.

Issue #88, opened 2022-01-12
Access to broadband


AmandaDoyle:
Description: Percent of households with access to broadband internet

Logic:

Source(s):ACS

Year(s):
- [ ]

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: 
Denominator:

SME:

Questions:

Roadblocks:

levysamu:
Years: 2015-19 ACS

Data is ACS PUMS and is already in PFF, do you need something more specific? If so, we'll ask Erica.

EricaMaurer:
Flagging issues from above...
Denominator is total households (not total population)
Should be referred to as estimates, not counts
All five columns of data should be included for each variable

PR #194, opened 2022-03-11
88 access broadband

td928:
## Overview

Use EDDT data shared by the Erica 

## load data
bulk of the work is actually taking place in the `load_clean_source_data` where the column names are renamed with a few dictionaries. It is important to note the sequence of the functions being run is relevant here because the race replacement must be the first. 

## access_broadband
this part of the function is simply pulling the right geography rows from the spreadsheet.


mbh329 (approved):
This all looks good, I tested it locally and the outputs and column names are good. I have one question - is the data reported at the household level? access_broadband_households suggests that it is but just curious because the numbers seem low but I also don't have a ref to how many households are in each borough/city/puma. This doesn't hold up anything as Erica calculated it for us

AmandaDoyle (approved):
Outputs are correct so I approve, but flagging the potential enhancements that Max pointed out

aggregate/quality_of_life/access_to_broadband.py
```diff
@@ -0,0 +1,84 @@
...
+        'sheet_name': "ACS15-19",
+        'usecols': "A, NM:QI",
+        "header": 0,
+        "nrows": 63
+    }
+
+    df = pd.read_excel(**read_excel_arg)
+
+    cols = [col.lower() for col in df.columns]
+    for code, race in race_suffix.items():
```
SashaWeinstein, line 76: Like this chunk of code, think it's readable

SashaWeinstein:
This all looks good except that the column names aren't intuitive to me. I see from the issue that we are to return 
Total households
With a computer
With broadband internet

"Access_broadband_households" to me reads like households with broadband internet. If we've already finalized then maybe it's not worth changing but could cause issues later on

SashaWeinstein (approved):
Approved with same comments as Max, borough mappers and puma clean should come from `PUMA_helpers`

SashaWeinstein:
lol wonderful Te pushed those changes the at the same time I made my comment

td928:
Thanks! To @mbh329 @SashaWeinstein comment, I incorporated those new functions. The new output for puma looks good as well. I was hesitant about the column naming as well except they are specified quite explicitly in the issue #88. @AmandaDoyle is the thinking here that we will be able to explain it in the data dictionary?

td928:
okay. New naming conventions updated. All output files were updated. @SashaWeinstein @AmandaDoyle @mbh329

SashaWeinstein (approved):
Great, these changes address all my concerns

PR #197, opened 2022-03-11
Export Demographics

SashaWeinstein:
# Export Demographics with ACS PUMS, Decennial Census and Census PUMS

This PR includes two different upgrades to the project. First there is Max's work to ingest Erica's census data and put it into EDDT format. Second there is code to clean up the output without changing any numbers. This includes work to reorder columns and changing column labels.  It's combined into one because I integrated Max's changes while also making changes to the ACS PUMS aggregation. 

## Decennial Census and 2000 Census PUMS

Max wrote this code, figuring out the regex to efficiently rename the column and coding up the mapping to our naming convention. This code can be used as a jumping-off point for other functions to ingest data from population. 

## Change column names in ACS PUMS Aggregation

The columns in the aggregated ACS PUMS data are named in the code in the `statistical/` folder. These are changes like "_count" being taken out, underscores converted to dashes, etc. 

## Refactor order columns

Code to reorder columns was taken out of the `PUMSAggregator` class and moved to a new 'aggregation_helpers.py' file. This code applied to ACS PUMS and census PUMS to ensure that they have the same column order. There are two  differences between the order column processes for ACS and census PUMS. The first is that denominator columns are labeled differently. The second is that age medians only have to be reordered for census PUMS as the order that the calculate code is called for ACS PUMS orders the columns correctly by default. 

## Quick fixes to external review PUMS

The version of external review PUMS relies on if/else statements that are inflexible. It makes sense to merge in this half-completed state as we won't use this code in the files we produce for the Apr 1st deadline. 

## Future upgrades

Some docstrings point to future upgrades that will happen in the more distant future

## Clean up internal review

Many files that we've already checked are deleted from internal review.


mbh329 (approved):
The logic looks good - dropped a few comments and a question but otherwise this all makes sense to me. A lot of the changes have to do with the column renaming from - to _, the code I wrote in the initial pull reqs and building out the collate steps to handle the new datasets. I think the refactoring of the `aggregate_PUMS` and moving those into the `aggregation_helpers.py` makes sense

aggregate/PUMS/count_PUMS_demographics.py
```diff
@@ -68,11 +68,11 @@ def LEP_assign(self, person):
 
     def age_bucket_assign(self, person):
         if person["AGEP"] < 16:
-            return "PopU16"
+            return "age_popu16"
```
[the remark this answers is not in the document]
SashaWeinstein: yep

external_review/external_review_PUMS.py
```diff
@@ -9,28 +9,97 @@
...
+            final = pd.concat([dec_census, ACS_PUMS], axis=1)
+
+    folder_path = f".staging/{eddt_category}"
+    if not path.exists(folder_path):
+        makedirs(folder_path)
+    final.to_csv(f".staging/{eddt_category}_{str(year)}_{geography}.csv")
+    return final
+
+
+def save_demographics(
```
td928, line 50: I see there is an inline comment about this should be deleted. @SashaWeinstein can confirm?

aggregate/decennial_census/census_2000_PUMS.py
```diff
@@ -0,0 +1,143 @@
...
+    Housing Security and Quality category indicators"""
+    df = df.drop(
+        df.filter(regex="P25pl|LTHS|HSGrd|SClgA|BchD|Occ|OOcc|ROcc").columns, axis=1
+    )
+    return df
+
+
+def remove_duplicate_cols(df):
+    """Excel spreadsheet has some duplicate columns that Erica used for calculations
+    - keep columns ending with Z for now"""
```
AmandaDoyle, line 58: Noting that we're always keeping the "Z" column

aggregate/housing_security/three_or_more_maintenance_deficiencies.py
```diff
@@ -3,7 +3,6 @@
 from ingest.load_data import load_HVS
 from aggregate.race_assign import HVS_race_assign
 from statistical.calculate_counts import calculate_counts
-from aggregate.clean_aggregated import sort_columns
 from ingest.clean_replicate_weights import rw_cols_clean
 from internal_review.set_internal_review_file import set_internal_review_files
```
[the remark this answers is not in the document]
td928: I think we now know we will take their data to redo this

tests/export/test_census_collate.py
```diff
@@ -0,0 +1,40 @@
...
+demos_2000 = save_PUMS(year="2000", **save_kwargs)
+demos_0812 = save_PUMS(year="0812", **save_kwargs)
+demos_1519 = save_PUMS(year="1519", **save_kwargs)
+
+
+def test_matching_column_orders_demographics_0812_1519():
+    assert (demos_0812.columns == demos_1519.columns).all()
+
+
+def test_matching_column_orders_demographics_2000_0812_1519():
```
td928, line 25: I really like those tests and we will need to do them again for the demographics "redo"

AmandaDoyle (approved):
I left one comment.  If I need to weigh in on any other comments please flag.

tests/export/test_census_collate.py
SashaWeinstein: ok nice, would love your input on future tests. I stopped testing my code when I was trying to hit our deadline for demographics last week and think that was a bad decision. I'm seeing that testing saves time even on short turnarounds

aggregate/decennial_census/census_2000_PUMS.py
SashaWeinstein: nice catch, removed that line in docstring

external_review/external_review_PUMS.py
SashaWeinstein: sorry what inline comment are you speaking of?

external_review/external_review_PUMS.py
```diff
@@ -9,28 +9,97 @@
...
+    eddt_category,
+    geography,
+    year,
+    test_data: bool = False,
+):
+    """Not used, can deleted if export works"""
```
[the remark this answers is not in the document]
td928: also it seems now it needs to merge in change from dev first

td928:
let me know if the way I changed it made sense. Thanks! @SashaWeinstein

SashaWeinstein:
I think those changes do make sense! I was going to make them but was unsure. Did you run `python3 -m external_review.external_review_PUMS demographics <year> <geography>` to check that it works? That is what I would to double check that nothing broke. Don't think it did but always want to be careful

td928:
Hey @SashaWeinstein I think I might have messed up my branch and struggling with resolving it now. Did the export worked on your end?

tests/export/test_census_collate.py
mbh329: Would also like to chat with you both on the tests, I think with Erica's data it should be a little easier to get the expected number of columns (each indicator should have 5 columns by race + 5 columns for the total population) so that might be a simple way to check out DO outputs

SashaWeinstein:
Hey sorry I missed this @td928. can help now

td928 (approved):
Hey I am going to approve and let this be merged in since the changes from dev looks good to me. Still can't test it on my end since I messed up by having my branch and remote diverged.

SashaWeinstein:
Ok neat, running tests on my machine. 

@mbh329's comments that we should check that outputted dataframes have exactly the number of columns we expect is very smart. Better to think about how many columns we expect (x indicators + y crosstabs * z columns for estimate, moe, etc) and check against that number

PR #204, opened 2022-03-15
Fix QOL Indicators: Broadband and Access to Transit: Car commute

mbh329:
This pull request covers some minor changes related to the reordering of columns in the QOL Access to broadband and access to transit: car commute indicators - there are functions  added to the `clean_aggregated.py ` and then the main accessor functions in the `access_ttransi_car.py` and `access_broadband.py` are updated.  There is also a change to the column naming conventions in the access to transit: car commute (from `access_carcommute_workers_` to `access_workers16pl_`) which the team thought was more intuitive and in line with other naming conventions for different indicators.


SashaWeinstein (commented):
Some of the changes here are mine so I can't sign off on those. The new column name for access car commute looks good. Is there a reason there aren't internal review files for access broadband as well?

mbh329:
good catch @SashaWeinstein, ill output the broadband internal review files now

td928 (approved):
new output looks good now @mbh329

AmandaDoyle (approved):

PR #206, opened 2022-03-16
Eviction cases indicator

SashaWeinstein:
# One-off Housing Security Indicator
Source data comes aggregated by community district. The only new thing is the three digit CD code. This code is the "borough code", a numeric representation of boroughs plus the typical CD number. For example manhattan CD 1 is 101, as Manhattan has borough code 1. 

## Refactor `community_district_to_PUMA`
Most of the existing code we use to map CD's to PUMA's is reusable here, we just change how the CD code is constructed. 

## Move function into `CD_helpers`
Some community district-specific code was moved to `CD_helpers`. 

## Outstanding Questions
I used the "eviction_filings" as the indicator name in the outputted data. This is easy to change if need be


utils/CD_helpers.py
```diff
@@ -1,8 +1,82 @@
...
+            else:
+                cd_code = construct_three_digit_CD_code(row["borough_code"], cd_num)
+            mapper[cd_code] = row.puma
+    df["puma"] = df[CD_col].replace(mapper)
+    return df
+
+
+def construct_three_digit_CD_code(borough_code: str, cd_num: str) -> str:
+
+    if len(cd_num) == 1:
```
td928, line 68: It took me so much time thinking about this for something else. I ended up going around this issue. Kudos to Sasha for tackling this and it seems to me now this should handle all different cases of the CDs

SashaWeinstein: this took me way longer than I thought it would. strangely difficult

PR #208, opened 2022-03-16
Expand Testing

SashaWeinstein:
# Two new "General Indicator Test" Files 
Four tests that were written for quality of life indicators were generalized for all indicators. The files with the new tests are in `tests/general_indicator_tests/`. Additionally`<category>_testing_helpers.py` files were added to organize accessor functions into iterables.

## Test Return Type
This tests that every accessors returns a dataframe. Some accessors were returning series which then can't be merged into the collated files for export

## Test Geographies
Three tests that ensure that when a particular geography is passed to an accessor, that accessor returns a dataframe with the correct index. 

## New testing action
A new `Test Indicators` workflow was added via `.github/workflows/test_indicators`. This runs the two tests described above. This workflow is set to run whenever any test or aggregation python code is changed so it will run on many pushes like the PUMS tests earlier. If this is helpful then great, if not let me know and I will roll it back to only test when we want to.

## Next steps
The key upgrade I think is having the tests and export process point to the same iterable of accessor functions. That way we don't have to worry about adding new accessor functions to both processes. Once that is done if the tests pass we will know with certainty that we won't run into issues with merging on indices in the export process. 
I've added this as a comment in the issue and can work on it tomorrow


SashaWeinstein:
Changes we talked about have been pushed. I did something slightly different than what we talked about: I put all the code to create the accessor lists in one file in `aggregate/all_acessors.py`. Think is is easier than having 4 different files that all do the same thing. The file is long but it's simple.

The external review process import from this new file.

Two new indicators were also changed to pass the tests. The indicators are eviction cases that I wrote and affordable units that Te wrote. The eviction cases was returning a series, and the affordable units was trying to import an outdated function. 

The workflow was updated to run whenever any aggregate, export, or test code is changed.

aggregate/all_accessors.py
```diff
@@ -0,0 +1,71 @@
...
+)
+
+# Housing Security imports
+# Currently count residential evictions is left out as it requires geocoding, can address this later if need be.
+# from aggregate.housing_security.three_or_more_maintenance_deficiencies import (
+#     count_units_three_or_more_deficiencies,
+# )
+
+# Three or more maintenance deficiences also left out as I'm not sure where we are getting data moving forward
+# from aggregate.housing_security.evictions_by_city_marshals import (
```
mbh329, line 17: This will be rectified with Te's pull request #211

aggregate/all_accessors.py
```diff
@@ -0,0 +1,71 @@
...
+
+census_accessors = [census_2000_pums, decennial_census_data]
+
+
+accessors = (
+    housing_security_accessors
+    + QOL_accessors
+    + housing_production_accessors
+    + census_accessors
+)
```
mbh329, line 71: I like keeping them all in one python script as per your comment...much cleaner and easier to find.

aggregate/housing_security/eviction_cases_housing_court.py
```diff
@@ -25,7 +25,7 @@ def load_filings():
 def eviction_cases(geography: str, write_to_internal_review=False):
     """Main Accessor"""
     filings = load_filings()
-    final = filings.groupby(geography).sum()["eviction_filings"]
+    final = filings.groupby(geography).sum()[["eviction_filings"]]
```
mbh329, line 28: the double brackets prevent `final` from being a series correct?

aggregate/housing_security/three_or_more_maintenance_deficiencies.py
```diff
@@ -41,8 +44,10 @@ def count_units_three_or_more_deficiencies(
...
         geo_col=geography_level,
         crosstab=crosstab,
+        keep_SE=False,
+        add_MOE=True,
     )
-    aggregated = sort_columns(aggregated)
+    # aggregated = sort_columns(aggregated)
     aggregated.name = (
         f"HVS_{year}_{geography_level}_crosstab_by_race_{crosstab_by_race}"
     )
```
mbh329, line 53: Once we confirm Te's most recent three or more maintenance deficiencies is correct, do we want to delete this file?

mbh329 (approved):
Looks good and is essentially what we talked about yesterday, I like keeping the accessor list in one python script - left some comments/questions but otherwise approved

aggregate/all_accessors.py
SashaWeinstein: yep!

aggregate/decennial_census/census_2000_PUMS.py
```diff
@@ -6,7 +6,7 @@
 """
```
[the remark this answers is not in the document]
SashaWeinstein: good flag, github says this conflict needs to addressed before merge. I will do no that now

aggregate/all_accessors.py
SashaWeinstein: ok nice glad you agree

aggregate/housing_security/eviction_cases_housing_court.py
SashaWeinstein: yea exactly, if you select a single column from a df with double brackets it returns a single column dataframe instead of a series

aggregate/housing_security/three_or_more_maintenance_deficiencies.py
SashaWeinstein: I think no. I added an exception that will throw an error if it's called but like the PUMS aggregation we want to keep it for future work

SashaWeinstein:
Hey @mbh329 can I get another review on this after merging in from dev? should be pretty simple but tired on friday afternoon seems like potential time for mistakes

mbh329 (commented):
The latest changes look good

PR #211, opened 2022-03-17
42 rent stablized three mainteinance

td928:
## Rent Stablisation and Three or More Mainteinances 

Based on the new spreadsheet from HPD for new rent stablisation and rework the three maintenance indicator.

## Functions
So the two functions for the two indicator and both take in geography as an input to produce the results. This work is really similar to the other housing security work. The only difference is where the column name change happens to accommodates the specificities of the source data. 

##PUMA_helpers
small changes to the puma helpers function to handle null cases in puma column 

## Review
Unfortunately this results do diverge quite significantly from the completed works from @SashaWeinstein based on the [borough level results in that branch](https://github.com/NYCPlanning/db-equitable-development-tool/blob/e1988df30729a92959654d8d37e618ea1ba67193/internal_review/HVS_2017_Borough_crosstab_by_race_False). I do want observe while the results are off by some factor between 2 to 3 but the trend in neighborhood seems to hold in Sasha's results


td928:
Sam just sent me the denominator files. Working on adding them now. 
[2017 HVS Denominator.xlsx](https://github.com/NYCPlanning/db-equitable-development-tool/files/8291580/2017.HVS.Denominator.xlsx)

AmandaDoyle:
> I would just assume my results are wrong? Is it worth digging into?

We're not short of work so leave it for now, but it will be something we'll likely troubleshoot together once it's confirmed that we will need to switch data sources from ACS to HVS and/or more than one year of data will need to be included in EDDT.

SashaWeinstein:
Ok neat. I know I said I was interested to see if my results were correct but at this point I'm more interested in getting the EDDT deliverables out the door and working through a new data product

td928:
Sorry for keep adding to this work. I realize the column needs reordering and also change the function name to something more explicit. It should be ready for final review now. 

Turns about in order to incorporate the other worksheet with denominator there is significant new work added and refactoring for the functions. Here are what changes were made. 

## extract_geography_dataframe
since I could not come up with a way to pre-join the denominator sheet and indicator sheet together before extract the results for different geography level. I decided that I would need to create two dataframes from the two sheets and rejoined them together. Another thing to note is the `denom` flag makes sure the right columns to drop depending processing which dataframe since denominator do not have Percent SE. 

## indicator functions
Now the two indcator functions simply rejoin the denom and indicator dataframes created and then start doing the final column change. No reordering for the columns at the moment but wonder if that is needed or this work. 

@SashaWeinstein @AmandaDoyle @mbh329

internal_review/housing_security/borough/units_rentstable.csv
```diff
@@ -0,0 +1,6 @@
+borough,units_rentstable,units_rentstable_moe,units_rentstable_cv,units_rentstable_pct,units_rentstable_pct_moe,units_occurental,units_occurental_moe,units_occurental_cv
```
mbh329, line 1: I think we want to round all the decimals (x.xx) but double check with @AmandaDoyle

aggregate/housing_security/rent_stable_three_mainteinance.py
```diff
@@ -0,0 +1,112 @@
...
+from typing import final
+import pandas as pd
+from utils.PUMA_helpers import clean_PUMAs
+from internal_review.set_internal_review_file import set_internal_review_files
+from aggregate.clean_aggregated import order_affordable
+from utils.PUMA_helpers import borough_name_mapper, clean_PUMAs
+
+suffix_mapper = {
+    "_N": "",
+    "Percent MOE\n(95% CI)": "pct_moe", # don't love this. But the order does matter here. As the MOE is a partial string match
```
SashaWeinstein: what don't you love about this?

mbh329 (commented):
This all works on my local machine and the outputs look accurate and follow expected patterns. two comments are that I think we want to round the decimals to two decimal places and there is a couple misspellings that should be corrected ("stablized" -> "stabilized", "mainteinance" -> "maintenance"). Otherwise looks good

SashaWeinstein:
It appears that the three or more maintenance deficiencies get it's denom info from the "2017 HVS Occupied Units" tab in the "2017 HVS Denominator.xlsx" file. That sheet doesn't say anything about maintenance deficiencies, it's tab names and headers talk about occupied units. Are sure this is the right source data for denoms of this indicator?

aggregate/housing_security/rent_stable_three_mainteinance.py
```diff
@@ -0,0 +1,112 @@
...
+    read_excel_arg ={
+        "io": "resources/housing_security/2017 HVS Denominator.xlsx",
+        "sheet_name": sheetname,
+        "header": 1,
+        "nrows": 63,
+        "usecols": "A:G"
+    }
+    data = pd.read_csv(**read_csv_arg)
+    data.columns = [c.replace(".1", "") for c in data.columns]
+    data.columns = ['SBA', 'PUMA', 'CD Name', 'N', 'SE', 'MOE\n(95% CI)', 'CV', 'Percent',
```
SashaWeinstein: Not sure what this line is doing, for both indicators it seems like the columns already have these labels? A better process is to rename from the original labels instead of hard-coding and re-naming I think

aggregate/housing_security/rent_stable_three_mainteinance.py
```diff
@@ -0,0 +1,112 @@
...
+    denom = pd.read_excel(**read_excel_arg)
+    denom["PUMA"] = denom["PUMA"].astype(str)
+    denom.columns = ['SBA', 'CD Name', 'PUMA', 'N', 'SE', 'MOE\n(95% CI)', 'CV',]
+
+    return data, denom
+
+def extract_geography_dataframe(clean_data: pd.DataFrame, geography: str, denom=False) -> pd.DataFrame:
+
+    if geography == "puma":
+        clean_data["puma"] = clean_data["PUMA"].apply(func=clean_PUMAs)
```
SashaWeinstein, line 118: This should go in load step instead of extract geography step I think

aggregate/housing_security/rent_stable_three_mainteinance.py
```diff
@@ -0,0 +1,112 @@
...
+    data.columns = ['SBA', 'PUMA', 'CD Name', 'N', 'SE', 'MOE\n(95% CI)', 'CV', 'Percent',
+       'Percent SE', 'Percent MOE\n(95% CI)']
+
+    denom = pd.read_excel(**read_excel_arg)
+    denom["PUMA"] = denom["PUMA"].astype(str)
+    denom.columns = ['SBA', 'CD Name', 'PUMA', 'N', 'SE', 'MOE\n(95% CI)', 'CV',]
+
+    return data, denom
+
+def extract_geography_dataframe(clean_data: pd.DataFrame, geography: str, denom=False) -> pd.DataFrame:
```
SashaWeinstein: I like breaking out the load and extract geography steps

SashaWeinstein (changes requested):
I have some questions/comments, but before addressing anything here I think the bigger issue is that the denom used for three or more maintenance deficiencies comes from a spreadsheet that doesn't mention maintenance deficiencies

mbh329:
@td928 @SashaWeinstein the data matrix mentions that occupied units should be the denominator

mbh329:
I'd like to take a look at this again on monday with a fresh set of eyes

mbh329:
Theres a couple other issues going on, did the cv and moe columns get mixed up?

aggregate/housing_security/rent_stable_three_mainteinance.py
td928: yeah this could be better rewritten it is actually a step to make sure the occupied rental units do not include the percentage MOE stuff so I would mess up the later extract step. Sort of hot fix for the issue and probably could be better written with drop?

aggregate/housing_security/rent_stable_three_maintenance.py
```diff
@@ -82,37 +80,22 @@ def load_source_clean_data(indicator: str) -> pd.DataFrame:
...
     return data, denom
 
-
-def extract_geography_dataframe(
-    clean_data: pd.DataFrame, geography: str, denom=False
-) -> pd.DataFrame:
+def extract_geography_dataframe(clean_data: pd.DataFrame, geography: str) -> pd.DataFrame:
 
     if geography == "puma":
         clean_data["puma"] = clean_data["PUMA"].apply(func=clean_PUMAs)
```
SashaWeinstein, line 101: I would still like to see this moved to `load_source_clean_data` as it's a cleaning step

aggregate/housing_security/rent_stable_three_maintenance.py
```diff
@@ -82,37 +80,22 @@ def load_source_clean_data(indicator: str) -> pd.DataFrame:
...
-    denom.columns = [
-        "SBA",
-        "CD Name",
-        "PUMA",
-        "N",
-        "SE",
-        "MOE\n(95% CI)",
-        "CV",
-    ]
+    if 'Percent' in denom.columns:
```
SashaWeinstein, line 91: Reading the [documentation on the drop method](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html), you can pass the `error` parameter as "ignore" and it will only drop the columns it finds, will take this block from four lines to  one

aggregate/housing_security/rent_stable_three_maintenance.py
```diff
@@ -7,23 +8,22 @@
 from utils.PUMA_helpers import borough_name_mapper, clean_PUMAs
 
 suffix_mapper = {
-    "_N": "",
+    "_N": "_count",
```
SashaWeinstein, line 11: I am all about adding the `_count` token here, but it needs to be added for variance of count as well. so the first three columns for three or more maintenance should be 

```
'units_threemaintenance_count', 'units_threemaintenance_count_cv', 'units_threemaintenance_count_moe'
```

SashaWeinstein (commented):
Ok nice this is much better! I have some substantial comments on getting the column label right and two suggestions on design/style. We do need to get the column labels right, the other two comments are not necessary

aggregate/housing_security/rent_stable_three_maintenance.py
td928: this would have downstream issues where the extra geographies columns need to be dropped. Can't come up with a really elegant solutions for this is why I keep it this way.

SashaWeinstein (approved):
Looks great!

SashaWeinstein (changes requested):
reorder column so moe comes before cv

SashaWeinstein:
hey sorry I didn't mean to push [b80ef4e](https://github.com/NYCPlanning/db-equitable-development-tool/pull/211/commits/b80ef4e5b4a94aa8f739c94d50fb4d5696e9d5fa), undid changes with second commit

td928:
Okay so no generalizable function was written because trying to handle both the reordering for denom and indicator itself it challenging with different measurements with the existing Frankenstein approach to the creating indicator in the first place.

SashaWeinstein (approved):
output looks good on machine

PR #215, opened 2022-03-21
Income restricted units

SashaWeinstein:
New indicator, nothing novel here. Number of NYCHA units and "RAD" units which refers to "Rental Assistance Demonstration". Data was prepped by HPD.

New column names are "NYCHA_units" and "RAD_units", these could be updated to reflect the year range (2014- present).

I wrote a new `filter_for_recognized_pumas` function that we can re-use in other indicators. Some places we hard-code number of rows in read csv/excel and I think it's better to filter for pumas in consistent way


mbh329 (approved):

internal_review/housing_security/borough/income_restricted_units.csv
```diff
@@ -0,0 +1,6 @@
+borough,NYCHA_units,RAD_units
+BK,54681.0,3940.0
+BX,40866.0,2458.0
```
mbh329, line 3: I don't think it makes sense to keep this as a float, should be an integer. I don't know if OSE cares but just want to flag

PR #218, opened 2022-03-22
Refactors: replace outdated function and standardize initializing empty dataframe

SashaWeinstein:
### Get rid of `add_leading_zero_PUMA` and implement `initialize_dataframe_geo_index`

Calls to `add_leading_zero_PUMA` were replaced were calls to `clean_PUMAs` as `clean_PUMAs` has the same functionality.

Calls of `pd.DataFrame` that initialize dataframes to hold agggregated info by PUMA/borough/citywide were replaced with calls to `initialize_dataframe_geo_index` which DRY's out this process


td928 (approved):
All look good from my end. Like the simplifying new design to create the empty dataframe. Just want to flag for you @SashaWeinstein in the pull request for #211. I have modified the puma_helpers. I don't think it would impact any of your logics here but let me know if you want to wait to test that on your first before merging.

PR #226, opened 2022-03-23
Add HSQ Indicators  (housing tenure) from 2000 Census PUMS

mbh329:
`pums_2000_hsq_housing_tenure.py`

Pretty straightforward script and very similar to the `2000_census_demographics` and `2000_census_economic`. Updated the tags to follow the newest conventions laid out in today's meeting with OSE.  Moved the global dictionaries to convert DCP population race codes and stat suffix mappers and the load data step  to the `utils/dcp_population_excel_helpers.py`. This script ONLY ingests/aggregates 2000 housing tenure PUMS data: `occupied_units, occupied_units_owner, occupied_units_renter`

`aggregation_helpers.py`

Updated this function to handle the newest tag conventions.

`utils/dcp_population_excel_helpers.py`
As mentioned above, added two global dictionaries `race_suffix_mapper` which converts the convention DCP population uses in the column names to indicate race to the race suffix used for EDDT and converts the stat convention used by dcp population to our conventions (i.e. `e` to `count`)

`internal_review`
Wrote files following what was laid out in the issue/epic for this indicator.

Things to check:

- [ ] Column order is correct
- [ ] naming conventions are followed correctly
- [ ] csv's are properly named


aggregate/aggregation_helpers.py
```diff
@@ -35,10 +35,10 @@ def order_aggregated_columns(
         if not household:
             for ind_category in categories[ind]:
                 for race_crosstab in categories["race"]:
-                    for measure in ["", "_pct"]:
+                    for measure in ["_count", "_pct"]:
                         column_label_base = f"{ind_category}_{race_crosstab}{measure}"
```
SashaWeinstein, line 39: this needs a refactor, race should come after measure and variance. I can make this fix as I know this function best

SashaWeinstein (changes requested):

aggregate/aggregation_helpers.py
SashaWeinstein: my apologies I don't think that feedback is correct, I got mixed up on where this token goes

SashaWeinstein (commented):
My apologies for the incorrect comments on race token. I pushed a commit that makes a some small fixes to column labels and order.  Otherwise everything else looks good, ran code on my machine

mbh329:
@SashaWeinstein Do I need to make any additional changes to the reordering? Based on your comments, I think everything is properly ordred? I am pushing a small change to a comment  in `dcp_population_excel_helpers.py` to avoid confusion but that will be my only change in the latest commit

mbh329:
I made a new commit to update a comment in the utils to avoid confusion

td928 (approved):

Issue #53, opened 2021-12-29
Housing units that are overcrowded


AmandaDoyle:
Description: Occupied units; units with >2 persons per bedroom (or more than 2 persons per studio)
There are two common overcrowding definitions used by HVS users: More than 2 persons per room (or more than two persons per studio) or more than two persons per bedroom (or more than two persons per studio). The latter is what conforms to occupancy standard so that is usually what HPD Housing Policy uses. 

Logic:

Source(s): NYC HVS

Year(s):
- [ ] 2002
- [ ] 2011
- [ ] 2017

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: Yes
Denominator:

SME: HPD

Questions:

Roadblocks:

levysamu:
Changes

Source: ACS
Years: 2000, 2008-12, 2015-19
SME: Pop Div

Issue #44, opened 2021-12-29
Rent-burdened households


AmandaDoyle:
Description: occupied units paying rent; gross rent as a percentage of household income is 30% or more.  Do not compare with 2000.

Logic:

Source(s): ACS PUMS

Year(s):
- [ ] 2008-2012
- [ ] 2015-2019

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: Yes
Denominator: occupied units paying rent

SME: DCP Population Division
Questions:

Roadblocks:
- 2008-2012 question

Issue #41, opened 2021-12-29
Housing tenure


AmandaDoyle:
Description: The number and percent of people who own their homes for three time periods.

Logic:

Source(s): Census, ACS

Year(s):
- [ ] 2000
- [ ] 2008-2012
- [ ] 2015-2019
Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: Yes
Denominator: total housing units

SME: DCP Population Division

Questions:

Roadblocks:
2008-2012

Issue #40, opened 2021-12-29
Median home value


AmandaDoyle:
Description: Median home value for two time periods inflated based on latest data.
Do not compare with 2000
Need distribution to calculate median
Inflate dollars based on latest dataset

Logic:

Source(s): ACS PUMS

Year(s):
- [ ] 2008-2012
- [ ] 2015-2019

Geographies:
- [ ] PUMA
- [ ] Borough
- [ ] City

Race and ethnicity breakdown: Yes
Denominator: DCP Population Division

SME:

Questions:
- what is the latest data we should use to inflate $?
Roadblocks:
- 2008 - 2012 PUMS issue

PR #227, opened 2022-03-23
225 housing security march 4 pop

td928:
## Overview
Six remaining indicattors (#41 #39 #40 #44 #199 #53)  from the March 4th POP data. At first glance a massive but hopefully become actually quite managable with explanations.

## Methodology
It is going to be similar to previous work to ingest population data with similar way to handle the columns. The work is broken out to smaller pieces of scripts that I thought make sense. There are two median indicators homevalue_median.py and rent_median.py which are single datapoint. Indicators share the same denominators are wrapped under the same functions units_overcrowd.py and units_housing_tenure.py and households_rent_burden.py each has three data points. 

## Review
The heavy lifting for this work is probably reviewing the final output but there are some patterns I want to share here. 

For sanity check on those results, I did some math on the number of columns and they are roughly the following logics: 

### three datapoint sets (housing tenure, overcrowd, rent burden)

(4 race + 1 total pop) * 3 data points * 5 measurements * 2 years = 150

### median single datapoint (homevalue and rent) 

(4race + 1 total pop) * 2 years * 3 measurements = 30 columns

## Column and Token Convention
the new output files should follow all the new column and token conventions that were established.


aggregate/clean_aggregated.py
```diff
@@ -4,6 +4,16 @@
 
 races = ["anh", "bnh", "hsp", "wnh"]
 
+reorder_mapper = {
```
SashaWeinstein, line 7: Makes sense to define as constant and load from helper module. Only comment is that it might be nicer to have a variable name that specifies that this mapper is for dcp population spreadsheets? Not necessary though

aggregate/clean_aggregated.py
```diff
@@ -47,4 +57,31 @@ def order_affordable(measures, income) -> List:
         for m in measures:
             rv.append(f"units_affordable_{i}{m}")
             
-    return rv
\ No newline at end of file
+    return rv
+
+
+def rename_col(df: pd.DataFrame, name_mapper: dict, race_mapper: dict, year_mapper: dict, suffix_mapper: dict):
```
SashaWeinstein, line 63: Similar comment here, could change function name to `rename_col_dcp_pop` or something. Really glad this process is being abstracted to a helper function!

aggregate/load_aggregated.py
```diff
@@ -69,3 +69,37 @@ def initialize_dataframe_geo_index(geography, columns=[]):
     rv = pd.DataFrame(index=indicies[geography], columns=columns)
     rv.index.rename(geography, inplace=True)
     return rv
+
+
+"""this is specifically to use for housing security and quality March 4th POP data"""
+def load_clean_pop_data(ind_name_str: str) -> pd.DataFrame:
```
SashaWeinstein, line 75: this function should def be renamed to `load_clean_QOL_pop_data` or something similar.

aggregate/housing_security/homevalue_median.py
```diff
@@ -0,0 +1,49 @@
...
+    final.set_index(geography, inplace=True)
+
+    final = rename_col(final, name_mapper, race_suffix_mapper, year_mapper, stat_suffix_mapper_md)
+
+    final.dropna(axis=1, how="all", inplace=True)
+    
+    if write_to_internal_review:
+        set_internal_review_files(
+            [(final, "homevalue_median.csv", geography)],
+            "housing_security",
```
mbh329, line 46: I think we can follow what Sasha and I did to reorder the columns for the data coming from dcp population - see the `pums_2000_hsq.py `script and `aggregation_helpers` code

aggregate/housing_security/households_rent_burden.py
```diff
@@ -0,0 +1,51 @@
+"""this chops up into six indicators they are """
+from typing import final
+import pandas as pd
+from aggregate.clean_aggregated import rename_col
+from utils.PUMA_helpers import clean_PUMAs, borough_name_mapper, get_all_boroughs, get_all_NYC_PUMAs
+from utils.dcp_population_excel_helpers import race_suffix_mapper, stat_suffix_mapper_ty
+from internal_review.set_internal_review_file import set_internal_review_files
+from aggregate.load_aggregated import load_clean_pop_data
+
+year_mapper = {"12": "0812", "19": "1519"}
```
mbh329, line 10: Can move this `year_mapper `to `dcp_population_excel_helpers.py`

aggregate/load_aggregated.py
SashaWeinstein: wait, I mean `load_clean_housing_security_dcp_pop`. Category here is housing security. What is docstring trying to say about QOL indicators?

aggregate/housing_security/homevalue_median.py
```diff
@@ -0,0 +1,49 @@
...
+
+year_mapper = {"12": "0812", "19": "1519"}
+
+def homevalue_median(geography: str, write_to_internal_review=False) -> pd.DataFrame:
+
+    name_mapper = {
+        "MdVl": "homevalue_median"
+    }
+
+    ind_name_str = "|".join([k for k in name_mapper.keys()])
```
mbh329, line 18: curious to see why you added this step but still haven't run it on my machine, just different than what I did

aggregate/housing_security/homevalue_median.py
```diff
@@ -0,0 +1,49 @@
+"""this chops up into six indicators they are """
```
SashaWeinstein, line 1: not sure what you are trying to say here?

aggregate/housing_security/homevalue_median.py
```diff
@@ -0,0 +1,49 @@
...
+import pandas as pd
+from aggregate.clean_aggregated import rename_col
+from utils.PUMA_helpers import clean_PUMAs, borough_name_mapper, get_all_boroughs, get_all_NYC_PUMAs
+from utils.dcp_population_excel_helpers import race_suffix_mapper, stat_suffix_mapper_md
+from internal_review.set_internal_review_file import set_internal_review_files
+from aggregate.load_aggregated import load_clean_pop_data
+
+year_mapper = {"12": "0812", "19": "1519"}
+
+def homevalue_median(geography: str, write_to_internal_review=False) -> pd.DataFrame:
```
SashaWeinstein, line 12: This function seems great, uses helper functions clearly

aggregate/housing_security/households_rent_burden.py
```diff
@@ -0,0 +1,51 @@
...
+
+    final = rename_col(final, name_mapper, race_suffix_mapper, year_mapper, stat_suffix_mapper_ty)
+
+    if write_to_internal_review:
+        set_internal_review_files(
+            [(final, "households_rent_burden.csv", geography)],
+            "housing_security",
+        )
+
+    return final
```
mbh329, line 51: This all looks good - I think the code is very very similar to the comments I made on the `homevalue_median.py `script above

aggregate/housing_security/rent_median.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if geography == "citywide":
+        final = clean_data.loc[clean_data["Geog"] == "citywide"].rename(columns={"Geog": "citywide"}).copy()
+    elif geography == "borough":
+        boros = ["BX", "BK", "MN", "QN", "SI"]
+        clean_data["Geog"] = clean_data["Geog"].map(borough_name_mapper, na_action="ignore")
+        final = (
+            clean_data.loc[clean_data["Geog"].isin(boros)]
+            .rename(columns={"Geog": "borough"}).copy()
+        )
```
mbh329, line 31: Eventually we should put this into `utils/dcp_population_excel_helpers.py` - we use this all the time

aggregate/housing_security/units_overcrowd.py
```diff
@@ -0,0 +1,52 @@
...
+
+year_mapper = {"12": "0812", "19": "1519"}
+
+
+def units_overcrowd(geography: str, write_to_internal_review=False) -> pd.DataFrame:
+
+    name_mapper = {
+        "OcR1p": "units_overcrowded",
+        "OcRU1": "units_notovercrowded",
+        "OcHU2": "units_occupied",
```
mbh329, line 19: Maybe the `units_xxxxx` should just be in one script? This is going to add a duplicate units_occupied column that we don't want if I understand the collate process correctly @SashaWeinstein

aggregate/housing_security/households_rent_burden.py
```diff
@@ -0,0 +1,51 @@
+"""this chops up into six indicators they are """
```
SashaWeinstein, line 1: don't need this here

aggregate/housing_security/households_rent_burden.py
```diff
@@ -0,0 +1,51 @@
...
+        "OHURt": "households_grapi", 
+    }
+
+    
+    ind_name_str = "|".join([k for k in name_mapper.keys()])
+    #print(ind_name_str)
+
+    clean_data = load_clean_pop_data(ind_name_str)
+
+    if geography == "citywide":
```
SashaWeinstein, line 26: this if/else block can be abstracted into it's own function I think. fine for this PR but for future a good refactor

aggregate/load_aggregated.py
```diff
@@ -69,3 +69,37 @@ def initialize_dataframe_geo_index(geography, columns=[]):
...
+
+    df_1519 = pd.read_excel(**read_excel_arg["1519"])
+
+    df = pd.merge(df_0812, df_1519, on="Geog", how="left")
+
+    df = df.filter(regex=ind_name_str + "|Geog")
+
+    df.loc[df["Geog"] == "NYC", "Geog"] = "citywide"
+    
+    return df
```
mbh329, line 105: This is a proper design question for the group, I feel like some of our functions are being put into a few different python files that we should consider consolidating? For example, I have a `load_pums_2000` function in the `utils/dcp_population_excel_helpers` which now just loads the census data that erica sent us for the 2000 census pums - the code above seems to be very similar to that and is also coming form erica. I don't have an answer @td928 @SashaWeinstein @AmandaDoyle

utils/dcp_population_excel_helpers.py
```diff
@@ -17,6 +17,31 @@
...
+    "_1519e": "_1519_count",
+    "_1519m": "_1519_count_moe",
+    "_1519c": "_1519_count_cv",
+    "_1519p": "_1519_pct",
+    "_1519z": "_1519_pct_moe",
+}
+
+stat_suffix_mapper_md = {
+    "_0812e": "_0812_median",
+    "_0812m": "_0812_median_moe",
```
mbh329, line 35: awesome!

aggregate/load_aggregated.py
SashaWeinstein, line 75: another comment here: the `ind_name_str` parameter should be renamed to have "regex" in the same, something like `ind_name_regex`.

aggregate/housing_security/homevalue_median.py
SashaWeinstein, line 18: the `ind_name_str` variable should be renamed to have "regex" in the same, something like `ind_name_regex`.

aggregate/load_aggregated.py
```diff
@@ -69,3 +69,37 @@ def initialize_dataframe_geo_index(geography, columns=[]):
     rv = pd.DataFrame(index=indicies[geography], columns=columns)
     rv.index.rename(geography, inplace=True)
     return rv
+
+
+"""this is specifically to use for housing security and quality March 4th POP data"""
+def load_clean_pop_data(ind_name_str: str) -> pd.DataFrame:
+    """Function to merge the two files for the QOL outputs and do some standard renaming. Because
+    these are QOL indicators they remain in the same csv output with columns indicating year"""
+
```
SashaWeinstein, line 78: Maybe the code to construct the regex from the mapper dictionary should live here? would DRY out all the places where we have `ind_name_str = "|".join([k for k in name_mapper.keys()])`. so this function would be passed a mapper instead of the regex string and construct the regex string here

SashaWeinstein (commented):
Wrote up some thoughts, submitting to give you a chance to read before continuing with my review

aggregate/housing_security/households_rent_burden.py
SashaWeinstein, line 51: column order here is off. Order for first 3 categories without race crosstabs is correct. But then should have all races for first category, then all races for second category, etc. Instead this does all 3 categories white non-hispanic, all 3 categories black non-hispanic, etc

aggregate/housing_security/rent_median.py
```diff
@@ -0,0 +1,50 @@
...
+
+    final.dropna(axis=1, how="all", inplace=True)
+    
+    if write_to_internal_review:
+        set_internal_review_files(
+            [(final, "rent_median.csv", geography)],
+            "housing_security",
+        )
+
+    return final
```
SashaWeinstein, line 50: output looks good except race crosstabs are in wrong order, should go in alphabetical order

SashaWeinstein (commented):
finished up my review

mbh329 (commented):
This all looks good to me and I am inclined to approve this pull req. all the changes I suggested are design related and I think we should briefly chat to see if this is worth doing now or when we aren't up against tight deadlines. One thing I did notice, that i think for my own sanity would be good to figure out is if I should move the files I did for HSQ from the `aggregate/PUMS/` into `housing_security` or vice versa. @AmandaDoyle @td928 @SashaWeinstein

aggregate/housing_security/units_overcrowd.py
SashaWeinstein: oh yea that's a good flag

aggregate/housing_security/rent_median.py
SashaWeinstein: yep

aggregate/load_aggregated.py
mbh329: I reviewed the outputs on my local machine - everything looked good (col_names with proper tags and sequence, expected # of columns in each output, etc) except for the column order which I think needs to be implemented

aggregate/housing_security/units_overcrowd.py
td928: Good question. I thought about this and check in the source spreadsheet that Pop provided two denomintor with different column names "OcHU2" & "OcHU1" I believe. On the other hand, I am not sure whether the data populated those two columns are also identical or not.

aggregate/clean_aggregated.py
```diff
@@ -44,7 +44,7 @@ def order_PUMS_QOL_multiple_years(categories, measures, years):
         for c in categories:
             for r in races:
                 for m in measures:
-                    rv.append(f"{c}_{r}{y}{m}")
+                    rv.append(f"{c}{y}_{r}{m}")
```
SashaWeinstein, line 47: nice, changes are clear. checked and this doesn't break the car access QOL indicator that also uses this function

aggregate/housing_security/homevalue_median.py
```diff
@@ -39,6 +39,14 @@ def homevalue_median(geography: str, write_to_internal_review=False) -> pd.DataF
...
 
     final.dropna(axis=1, how="all", inplace=True)
+
+    col_order = order_PUMS_QOL_multiple_years(
+        categories=["homevalue_median"],
+        measures=["_median", "_median_moe", "_median_cv"],
+        years=["_0812", "_1519"],
+    )
+
+    final = final.reindex(columns=col_order)
```
SashaWeinstein, line 49: this one looks good to me!

aggregate/housing_security/rent_median.py
```diff
@@ -40,6 +40,15 @@ def rent_median(geography: str, write_to_internal_review=False) -> pd.DataFrame:
...
 
     final.dropna(axis=1, how="all", inplace=True)
+
+    col_order = order_PUMS_QOL_multiple_years(
+        categories=["rent_median"],
+        measures=["_median", "_median_moe", "_median_cv"],
+        years=["_0812", "_1519"],
+    )
+
+    final = final.reindex(columns=col_order)
```
SashaWeinstein, line 50: rent median looks good

SashaWeinstein (changes requested):
The new function seems to work where it is applied. It needs to be applied to rent burden, units housing tenure as well right?

td928:
alright I think looking at the checklist they are all done. Avoid some more intense refactoring to save time. Let me know your thoughts! Thank you! @SashaWeinstein @mbh329

aggregate/aggregation_helpers.py
```diff
@@ -89,3 +89,22 @@ def get_category(indicator, data=None):
             categories.remove(None)
         categories.sort()
         return categories
+
+def get_geography_housing_security_pop_data(clean_data: pd.DataFrame, geography:str):
```
SashaWeinstein, line 93: this looks great!

aggregate/housing_security/units_overcrowd.py
SashaWeinstein: I think they are, this is issue we've seen before with dcp population data

SashaWeinstein (commented):
Ran code on my machine and everything looks good! Last thing is the duplicate units occupied columns, my vote is to keep them in the units housing tenure indicator as it's first in alphabet

SashaWeinstein (approved):
Looks good to me! ran a quick merge on my machine and didn't get duplicate columns

PR #229, opened 2022-03-24
Refactor 2000 HSQ Housing Tenure 

mbh329:
This pull req covers requests discussed between @td928, @SashaWeinstein and myself:

- add 2000 to the column name for the housing security quality data (housing tenure data) for the 2000 pums data 
- move `pums_2000_hsq_housing_tenure.py` to the `aggregate/housing_security` folder
- add new accessor 
- write updated internal review


SashaWeinstein (approved):
Looks good. Tests pass. No need to re-write internal review files I think, those are just used to check results when indicator code is reviewed

AmandaDoyle (approved):

PR #231, opened 2022-03-24
Renaming and Adjusting Housing Production Columns

td928:
## Overview

very simple work to mostly add _count token to housing production indicators (only three: change in units, historic districts and hpd affordable housing). Also a quickly reordering columns for the historic district (only three columns) so the denom comes last.


SashaWeinstein:
The big issue here is that the branch you worked off is old and way behind `dev`. The general indicator tests aren't on this branch. Can you merge in `dev`? Then I will review

SashaWeinstein:
The branch with my work on this improvement is `223-Count-Token-Housing-Prod`, which is more up to date

td928:
> The big issue here is that the branch you worked off is old and way behind `dev`. The general indicator tests aren't on this branch. Can you merge in `dev`? Then I will review

Hey Sasha not sure what you said was true. I merged in dev before starting on this. And from only 10 files change compared to dev I think it seems merge was successful.

SashaWeinstein:
sorry I'm a dummy I forgot to pull after checking out the feature branch

aggregate/housing_production/area_within_historic_district.py
```diff
@@ -31,7 +37,8 @@ def fraction_historic(geography_level):
     gdf[["area_historic_pct", "area_historic_sqmiles"]] = gdf.apply(
```
SashaWeinstein, line 37: my solution was just to change the second string in this line to `area_historic_sqmiles_count`

aggregate/housing_production/area_within_historic_district.py
```diff
@@ -22,6 +23,11 @@ def area_historic_internal_review():
         "housing_production",
     )
 
+def rename_col(cols) -> List:
```
SashaWeinstein, line 26: this is fine but it introduces a lot of complexity for two column labels to be changed. In my solution I just changed the lines where the new columns were set

aggregate/housing_production/change_in_units.py
```diff
@@ -149,6 +150,11 @@ def clean_jobs(df):
 
     return df
 
+def rename_col(cols) -> List:
```
SashaWeinstein, line 153: This seems like a fine solution for this indicator. I changed the strings in the rename call in `pivot_and_flatten_index` as I think it's nice to have the entire indicator label hard-coded somewhere but this is fine as well

td928:
> Ok I was wrong about the branch being out of date but I had already done some of this work on ` 223-Count-Token-Housing-Prod`. I thought this was clear from context as all the issues this improvement were made around the time. We can take your solution but we need to be more careful about this in the future

yeah I should pay more attention to the issue numbers. But I was thrown off by the default git fetch did not return 223-Count-Token-Housing-Prod branch so I went to the most plausible one next in line. Probably should ask you first.

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
```diff
@@ -85,7 +85,7 @@ def pivot_and_flatten_index(df, geography):
 
     cols = [c.replace("new_construction", "newconstruction") for c in cols]
 
-    df.columns = cols
+    df.columns = [col + "_count" for col in cols]
```
SashaWeinstein, line 88: does it make sense to combine this with line 86? So line 86 would be 

```python
cols = [c.replace("new_construction", "newconstruction") + "_count" for c in cols]
```

SashaWeinstein (approved):
This all looks good, I had different ideas about what was clearest but this is fine the way it is. I left one suggestion but it's a design issue so you can take it or leave it

SashaWeinstein:
> > Ok I was wrong about the branch being out of date but I had already done some of this work on ` 223-Count-Token-Housing-Prod`. I thought this was clear from context as all the issues this improvement were made around the time. We can take your solution but we need to be more careful about this in the future
> 
> yeah I should pay more attention to the issue numbers. But I was thrown off by the default git fetch did not return 223-Count-Token-Housing-Prod branch so I went to the most plausible one next in line. Probably should ask you first.

ok yes I see I never pushed that branch. so definitely my fault, I assumed that it was pushed and that you knew I had started on this improvement but neither of those things were true

aggregate/housing_production/change_in_units.py
td928: yeah the issue here is the denom then get left out without the token. I don't see a clever way to add that for the moment

td928:
yeah I can l do a better job communicating next time! Took your suggestion for list comprehension change and test it the outlook look good.

SashaWeinstein (approved):
great, glad that suggestion was helpful. think the new code is readable enough

Issue #233, opened 2022-03-24
Income-restricted housing units - rename columns


AmandaDoyle:
Since we already need to regenerate for DHS.  One more nitpicky issue for Income-restricted housing units

change
`rad_units_count` to `units_rad_count`
`nycha_units_count` to `units_nycha_count`

Issue #221, opened 2022-03-22
Add `_count` token to housing security and quality


td928:
breakdown the #217 to smaller pieces to meet our delivery time

td928:
this work is in progress #224

PR #234, opened 2022-03-24
Update income_restricted_units.py

AmandaDoyle:
#233 
Update columns to be consistent units_ as subcategory


SashaWeinstein (approved):
This looks good to me. Code works in ipython and tests pass. If units is housing security subcategory, what are the other housing security sub-categories?

td928:
Hey @SashaWeinstein just want to also flag this overlap with your work on branch #221 with the column name changes

PR #240, opened 2022-03-25
Fix borough level aggregation for two housing security indicators

SashaWeinstein:
## Issues with missing info at borough level for eviction cases and evictions by city marshals

Branch was opened to just fix eviction cases and I saw issue with evictions when checking

### Eviction cases
Issue was that conversion from numeric borough code was replacing with text of full borough name. This was fixed in `CD_helpers.get_borough_num_mapper` to set text borough as code ("BX", "MN") instead of full borough name. 

### Evictions by City Marshals
Was missing staten island evictions because boroughs are all uppercase in source data and my code to clean converted all but first letter to lowercase. This however sent "STATEN ISLAND" to "Staten island" and the lowercase "i" kept those records from being converted to "SI". This was hot-fixed, a more elegant solution would involve a case-insensitive match

### DIFF
--- a/aggregate/housing_security/evictions_by_city_marshals.py
+++ b/aggregate/housing_security/evictions_by_city_marshals.py
@@ -1,13 +1,21 @@
 import pandas as pd
+from internal_review.set_internal_review_file import set_internal_review_files
 from utils.PUMA_helpers import assign_PUMA_col, clean_PUMAs, borough_name_mapper
 
 
-def count_residential_evictions(geography_level, debug=False):
+def count_residential_evictions(
+    geography_level, write_to_internal_review=False, debug=False
+):
     """Main accessor of indicator"""
     residential_evictions = load_residential_evictions(debug)
     aggregated_by_geography = aggregate_by_geography(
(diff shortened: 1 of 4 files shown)


mbh329 (approved):

aggregate/housing_security/evictions_by_city_marshals.py
```diff
@@ -22,6 +30,11 @@ def load_residential_evictions(debug) -> pd.DataFrame:
         residential_evictions["borough"].str[0]
         + residential_evictions["borough"].str[1:].str.lower()
     )
+
+    # Hot fix, can be made neater
+    residential_evictions["borough_name"] = residential_evictions[
+        "borough_name"
+    ].replace({"Staten island": "Staten Island"})
```
AmandaDoyle, line 37: Got it.  This is the not elegant solution.  One idea for future improvement is to cast the value to all upper or lower case and then do a value match
i.e. set Staten Island to be "STATEN ISLAND" then map that value.

AmandaDoyle (approved):

PR #246, opened 2022-03-25
Fix subgroup order in access_transit_car

mbh329:
Order of the year and subgroup were incorrect in access_transit_car which populated NAN values when the columns were reordered, fix that and write new csvs for review


SashaWeinstein (approved):
Tried it out on my machine and everything looks good to me

AmandaDoyle (approved):

PR #249, opened 2022-03-25
Fix tests and remove unused imports 

SashaWeinstein:
Minor changes that shouldn't impact any functionality. 

Tests now run for all categories

Unused imports cleared from 3 categories we've exported so far


td928 (approved):
Okay glad I checked this. Seems I can use those new changes for the test I am about to do.

AmandaDoyle (approved):

PR #252, opened 2022-03-28
ACS PUMS economics from DCP population

SashaWeinstein:
# Economic indicators for 2008-2012 and 2015-2019
Data comes from DCP population in spreadsheets formatted in similar way to their data for demographics. We have a pretty well-established process for transforming this type of data. 

My code to rename cols is slightly different than what's in the demographics code. I wrote a `convert_col_label` script that gets called for each column label. I also wrote new mappers as the existing ones either include underscores or are specific to certain years. I prefer to add underscores via the logic in the function rather than get them from a mapper


mbh329:
Aren't these indicators "demographic"?

SashaWeinstein:
# Add medians and Re-order columns
I opened up this PR when the work was 50% done, that was my bad. 

## Add medians
Adding the medians was pretty simple, median wages have leading "MW" in indicator label. Curious that source data doesn't have median wages for occupations crosstabbed by race. It does have median wages by industry crosstabbed by race.


## Re-order coumns
The existing `order_aggregated_columns` worked for the count indicators. It's not exactly right as we would rather have each occupation act as an indicator rather than a category. The way this orders the columns all occupations are together and then all crosstabs on all occupations follow. It would be better to put each occupation crosstab next to the numbers for all races of that occupation but it's ok for now.

I had to write new code for the medians re-order as one indicator is not crosstabbed by race which is an unusual pattern. 

I checked the math for the expected number of columns and have assertions that all cols are preserved when we reindex.
 
## Future work
Lots of good refactoring to do for the reordering columns and dcp population processing in general. The assertions to make sure columns aren't dropped is a good QAQC measure. 

## Unsure
I set the indicator label for income bands as `households_<income band>`. Easy to change if we want something different

aggregate/PUMS/pums_0812_1519_economics.py
```diff
@@ -0,0 +1,228 @@
...
+from internal_review.set_internal_review_file import set_internal_review_files
+from utils.PUMA_helpers import borough_name_mapper, clean_PUMAs, get_all_boroughs
+from utils.dcp_population_excel_helpers import (
+    race_suffix_mapper_global,
+    count_suffix_mapper_global,
+    median_suffix_mapper_global,
+    remove_duplicate_cols,
+)
+
+occupations = ["mbsa", "srvc", "slsoff", "cstmnt", "prdtrn"]
```
mbh329, line 12: I agree that the mappers shouldn't have the underscores - I prefer your method of doing the mappers without underscores as it takes away from some of the confusion as to where in the mapper the underscore goes when it replaces the original

mbh329: on refactor, we can update the other scripts

aggregate/PUMS/pums_2000_demographics.py
```diff
@@ -53,12 +54,6 @@ def filter_to_demo_indicators(df):
     return df
```
mbh329, line 56: Thanks for moving this, just have to make sure I update this on my end

aggregate/all_accessors.py
```diff
@@ -98,6 +99,7 @@
 census_accessors = [
     census_2000_pums_demographics,
     decennial_census_data,
+    ACS_PUMS_economics,
```
mbh329, line 102: @SashaWeinstein @td928 Do we wanna rename the accessor to match the file name?

internal_review/economics/PUMA/ACS_PUMS_economics_0812.csv
```diff
@@ -0,0 +1,56 @@
...
...8,,,49501,6201,7.6,,,46612,3714,4.8,,,24455,7334,18.2,,,29012,5592,11.7,,,76335,14648,11.7,,,15401,1104,4.4,100,,97639.0,,,,,29827,27871.0,56.8,,,66875.0,31244.0,28.4,,,64720.0,68795.0,64.6,,,45533.0,9772.0,13.0,,,74440,24132.0,19.7,,,81751.0,24275.0,18.1,,,73015.0,17068.0,14.2,,,62583,18934.0,18.4,,,69955,7425.0,6.5,,,26445.0,11897.0,27.3,,,48795,22906.0,28.5,,,101162.0,22043.0,13.2,,,6333,717,6.9,100,,,,,,,75321.0,,,,,68884.0,63997.0,56.5,,,51106.0,,,,,16178,16320.0,61.3,,,57568.0,13690.0,14.5,,,91299.0,26380.0,17.6,,,43044,11104.0,15.7,,,42252.0,22388.0,32.2,,,43364,4840,6.8,,,36577.0,8956.0,14.9,,,29756.0,5755.0,11.8,,,58630.0,17931.0,18.6,,,20249,1230,3.7,100,,30685.0,,,,,29158.0,13961.0,29.1,,,29400,8171,16.9,,,37034.0,12439.0,20.4,,,19635,9438.0,29.2,,,32663,7870,14.6,,,46505.0,13587.0,17.8,,,50524,12820,15.4,,,39358,13618,21.0,,,30544,6050,12.0,,,20250,4388,13.2,,,14984,11831.0,48.0,,,62975,19779.0,19.1,,,2716,439,9.8,100,,,,,,,19868.0,19846.0,60.7,,,31775.0,122955.0,235.2,,,58911.0,53644.0,55.4,,,45268.0,90909.0,122.1,,,11000.0,,,,,91536.0,48657.0,32.3,,,43990.0,89289.0,123.4,,,70114.0,65402.0,56.7,,,67518,25593,23.0,,,33764,40233.0,72.4,,,41147,43207.0,63.8,,,125536.0,,,,
```
mbh329, line 2: I think the educational attainment indicators are off from the 2000 census pums economic column names. These are the original col and the renamed cols used in the mapper: 
```
    "lths": "edu_lths",
    "hsgrd": "edu_hsgrd",
    "sclga": "edu_sclga",
    "bchd": "edu_bchd",
```

aggregate/all_accessors.py
SashaWeinstein: yea I think the ACS PUMS vs 2000 naming convention is concise and sufficiently specific. Different than the QOL, housing security and housing prod categories in that there are fewer accessors and so convention matters a little less. Te we do need to figure out whether we want upper or lowercase though

mbh329: That works

aggregate/PUMS/pums_0812_1519_economics.py
```diff
@@ -0,0 +1,228 @@
...
+    "trwhu",
+    "info",
+    "fire",
+    "pfmg",
+    "edhlt",
+    "arten",
+    "oth",
+    "pbadm",
+]
+ages = ["p25p", "p16t64"]
```
td928, line 29: Hey Sasha looking at the [data dictionary](https://nyco365.sharepoint.com/:x:/s/NYCPLANNING/civic-engagement-studio/EVBcBM4HnJFCtab3bRKNOxgBSTVR9j0wlk-D4uM1GxyNhw?e=NpCgRN). Should the age_p25p be age_p25pl?

SashaWeinstein: good catch, can change that now

internal_review/economics/citywide/ACS_PUMS_economics_0812.csv
```diff
@@ -0,0 +1,2 @@
...gff_anh_count_moe,mwagff_anh_count_cv,mwagff_anh_pct,mwagff_anh_pct_moe,mwcnstn_anh_count,mwcnstn_anh_count_moe,mwcnstn_anh_count_cv,mwcnstn_anh_pct,mwcnstn_anh_pct_moe,mwmnfct_anh_count,mwmnfct_anh_count_moe,mwmnfct_anh_count_cv,mwmnfct_anh_pct,mwmnfct_anh_pct_moe,mwwhlsl_anh_count,mwwhlsl_anh_count_moe,mwwhlsl_anh_count_cv,mwwhlsl_anh_pct,mwwhlsl_anh_pct_moe,mwrtl_anh_count,mwrtl_anh_count_moe,mwrtl_anh_count_cv,mwrtl_anh_pct,mwrtl_anh_pct_moe,mwtrwhu_anh_count,mwtrwhu_anh_count_moe,mwtrwhu_anh_count_cv,mwtrwhu_anh_pct,mwtrwhu_anh_pct_moe,mwinfo_anh_count,mwinfo_anh_count_moe,mwinfo_anh_count_cv,mwinfo_anh_pct,mwinfo_anh_pct_moe,mwfire_anh_count,mwfire_anh_count_moe,mwfire_anh_count_cv,mwfire_anh_pct,mwfire_anh_pct_moe,mwpfmg_anh_count,mwpfmg_anh_count_moe,mwpfmg_anh_count_cv,mwpfmg_anh_pct,mwpfmg_anh_pct_moe,mwedhlt_anh_count,mwedhlt_anh_count_moe,mwedhlt_anh_count_cv,mwedhlt_anh_pct,mwedhlt_anh_pct_moe,mwarten_anh_count,mwarten_anh_count_moe,mwarten_anh_count_cv,mwarten_anh_pct,mwarten_anh_pct_moe,mwoth_anh_count,mwoth_anh_count_moe,mwoth_anh_count_cv,mwoth_anh_pct,mwoth_anh_pct_moe,mwpbadm_anh_count,mwpbadm_anh_count_moe,mwpbadm_anh_count_cv,mwpbadm_anh_pct,mwpbadm_anh_pct_moe
```
mbh329, line 1: @SashaWeinstein  the code has the changes with the `edu_` so I am assuming you didn't write new internal review outputs correct?

aggregate/aggregation_helpers.py
```diff
@@ -19,12 +19,12 @@ def order_aggregated_columns(
     df: pd.DataFrame,
     indicators_denom,
     categories,
-    household,
-    census_PUMS=False,
+    household=False,
+    exclude_denom=False,
```
td928, line 23: this would have bunch of downstream effects on the demographics work. Need to sort out depending which branch gets merged first.

SashaWeinstein: renaming the parameter in the function also auto-renames the parameter passed to the function because VScode re-names parameters across files. This definitely shouldn't break anything.

I also think setting a default for household is nice and shouldn't break any other accessors?

mbh329:
I think this all looks good - will wait for @AmandaDoyle on the column names for the `edu_` indicators to approve

aggregate/aggregation_helpers.py
td928: yeah although for the work currently unavailable in the branch is not the case right? I think for the demographics stuff still in the PR probably might need an extra look at. But overall I feel like merging this PR from you would be easier first step.

mbh329 (approved):

td928 (approved):

td928:
Hey Sasha immediately after approval I remember I had this comments

SashaWeinstein:
Hey Te what comments?

td928:
sorry just updated above

SashaWeinstein:
sorry I still don't understand lol. Good to merge?

td928:
this is just awful fumble from me but I was talking about this thread https://github.com/NYCPlanning/db-equitable-development-tool/pull/252#issuecomment-1082182028

td928 (approved):

PR #253, opened 2022-03-28
241 demographics pums 0812 1519

td928:
## Overview
this work is used to incorporate the two spreadsheets (EDDT_Dem_ACS2008-2012.xlsx and EDDT_Dem_ACS2015-2019.xlsx data) from Pop for demographics indicators. This work changes a lot since now I realize the output can be for each year range and not all under the same files. 

## acs_pums_demographics
this is the main accessor. Trying to take advantages of the existing work. The concepts is very similar to ingest PUMS data from Pop and the process to rename columns and then reorder them are also what is taking place inside this function.

## rename_columns_demo
One distinction compared to the past work for this is since there two sets of columns for each year. So a year is passed onto this function. 

## order_aggregate_columns
it's the same function used to reorder the columns for PUMS demoraphics indicators before


aggregate/PUMS/pums_0812_1519_demographics.py
```diff
@@ -0,0 +1,50 @@
+"""This script takes the xlsx files Erica from DCP Population sent us imports them,
```
mbh329, line 1: There are columns in the final dataframes that are missing. In the source data, erica provides columns `pop3_12/pop3_19` and `pop1_12/pop1_19` which both refer to total population (the denominator for the foreign born poulation and then the denominator for age). Neither of these make it into the final data that is accessed via the main accessor `acs_pums_demographics`. I believe these are duplicates but weren't caught because they are unique column names. First step is to drop one of them, update them to follow convention (i.e. `pop1_12e` : `pop_0812_count`, etc), and then include them in the final output. the reason they aren't making it into the final dataframe is the reorder function, I think you need to include pop in the categories for it to be captured in the final output

aggregate/PUMS/pums_0812_1519_demographics.py
```diff
@@ -0,0 +1,50 @@
...
+from utils.PUMA_helpers import dcp_pop_races
+from internal_review.set_internal_review_file import set_internal_review_files
+from aggregate.load_aggregated import load_clean_pop_demographics
+
+endyear_mapper = {
+    "0812": "12", 
+    "1519": "19" 
+}
+
+def acs_pums_demographics(geography: str, year: str, write_to_internal_review=False) -> pd.DataFrame:
```
mbh329, line 20: update accessor function name to match script

aggregate/aggregation_helpers.py
```diff
@@ -1,4 +1,8 @@
+from matplotlib.pyplot import axis
```
mbh329, line 1: @SashaWeinstein @td928 We rneed to organize our helper functions and scripts better but can wait till refactor/not up against deadlines

mbh329: lets discuss where this should all live

aggregate/clean_aggregated.py
```diff
@@ -4,7 +4,17 @@
 
 races = ["anh", "bnh", "hsp", "wnh"]
 
-reorder_mapper = {
+demo_suffix = {
+    ## Rename the demographic race columns with wiki conventions
```
mbh329, line 8: TODO: reorg aggregation helper functions/excel helper mappers

internal_review/demographics/borough/ACS_PUMS_demographics_0812.csv
```diff
@@ -0,0 +1,6 @@
...t64_anh_count_moe,age_p16t64_anh_count_cv,age_p16t64_anh_pct,age_p16t64_anh_pct_moe,age_p16t64_bnh_count,age_p16t64_bnh_count_moe,age_p16t64_bnh_count_cv,age_p16t64_bnh_pct,age_p16t64_bnh_pct_moe,age_p16t64_hsp_count,age_p16t64_hsp_count_moe,age_p16t64_hsp_count_cv,age_p16t64_hsp_pct,age_p16t64_hsp_pct_moe,age_p16t64_wnh_count,age_p16t64_wnh_count_moe,age_p16t64_wnh_count_cv,age_p16t64_wnh_pct,age_p16t64_wnh_pct_moe,age_p65pl_anh_count,age_p65pl_anh_count_moe,age_p65pl_anh_count_cv,age_p65pl_anh_pct,age_p65pl_anh_pct_moe,age_p65pl_bnh_count,age_p65pl_bnh_count_moe,age_p65pl_bnh_count_cv,age_p65pl_bnh_pct,age_p65pl_bnh_pct_moe,age_p65pl_hsp_count,age_p65pl_hsp_count_moe,age_p65pl_hsp_count_cv,age_p65pl_hsp_pct,age_p65pl_hsp_pct_moe,age_p65pl_wnh_count,age_p65pl_wnh_count_moe,age_p65pl_wnh_count_cv,age_p65pl_wnh_pct,age_p65pl_wnh_pct_moe,age_median_median,age_median_median_moe,age_median_median_cv,age_median_anh_median,age_median_anh_median_moe,age_median_anh_median_cv,age_median_bnh_median,age_median_bnh_median_moe,age_median_bnh_median_cv,age_median_hsp_median,age_median_hsp_median_moe,age_median_hsp_median_cv,age_median_wnh_median,age_median_wnh_median_moe,age_median_wnh_median_cv
```
mbh329, line 1: output column names look good but see first comment, missing population columns in all outputs

aggregate/PUMS/pums_0812_1519_demographics.py
td928: got it. I only include the columns that were included in the earlier demographics work probably was a mistake. So to reiterate what you said so I have it correct, I will only need only include one set i.e. `pop1` and for both year and convert to convert to our naming convention to get the correct output

mbh329: Yeah it was a mistake on my end as well - the issue is with the reordering

td928:
took way longer than i wanted to but finally got it to work. Lessons learned: our existing flow to do this organizationally is definitely a little messy that needs refactoring. 

## Age Over 5 and Total Population
this was not in the PUMS 2000 dem work but now is added to this work and should also be incorporated to add to the PUMS 2000 dem before that work is finalized. @mbh329 

## Pytest
the new tests actually don't work as they are constructed. I don't know if we will have enough time for me to rewrite them but basically now the test takes only a geography parameter but for any of the demographics and economics work it will need both year and geography. Not the heavies lift and maybe @SashaWeinstein is already making ways down that road?

internal_review/demographics/borough/ACS_PUMS_demographics_0812.csv
mbh329: I mean the data points for the denominators for foreign-born population pct and the age group pct

AmandaDoyle: ah I see.  Apologies.  Yes, those need to be included.  Thanks for flagging.

mbh329:
@td928 What do you mean by the work wasn't included in the 2000 pums demographics and need to incorporate it? I think the age_5pl work is in the demographic ouputs, just missing the pop columns although @AmandaDoyle made a comment that the outputs look good. I will start running this branch on my local

tests/general_indicator_tests/general_indicator_test_helpers.py
```diff
@@ -3,9 +3,14 @@
 accessors = Accessors()
 
 
-def get_by_geo():
+def get_by_geo(housing_security=False, pums_demographics=False):
     """housing security parameter is temporary"""
-    accessors_list = accessors.all
+    if housing_security:
```
SashaWeinstein, line 8: I think that in `dev` get_by_geo shouldn't have any parameters. Tests on particular categories for developments should stay on local branches and not be merged in

aggregate/PUMS/pums_0812_1519_demographics.py
```diff
@@ -33,7 +33,6 @@ def acs_pums_demographics(geography: str, year: str, write_to_internal_review=Fa
         census_PUMS=True,
         demographics_category=True,
     )
-    final.dropna(axis=1, how="all", inplace=True)
```
SashaWeinstein: Issue with extra `age_p5pl` being added. Think it's this block of code

```python
if census_PUMS and ind == "LEP":
    col_order.append("age_p5pl")
```
I think this can come out entirely? Along with 

```python 
if not census_PUMS:
    col_order.append(f"{ind_category}_pct_denom")
```

SashaWeinstein:
Ok shoot I ran some tests and changes here break the `census_2000_pums_demographics` function. I think @td928 and @mbh329  you should work this out as you two worked on this code? 
Stack trace I got looks like this:
```python
tests/general_indicator_tests/test_tokens.py:5: in <module>
    by_puma, by_borough, by_citywide = get_by_geo(pums_demographics=True)
tests/general_indicator_tests/general_indicator_test_helpers.py:14: in get_by_geo
    by_puma.append((a("puma"), a.__name__))
aggregate/PUMS/pums_2000_demographics.py:131: in census_2000_pums_demographics
    final = order_pums_2000_demographics(final)
aggregate/PUMS/pums_2000_demographics.py:144: in order_pums_2000_demographics
    final = order_aggregated_columns(
aggregate/aggregation_helpers.py:35: in order_aggregated_columns
    for ind_category in categories[ind]:
E   KeyError: 'total_pop'
```

aggregate/PUMS/pums_0812_1519_demographics.py
td928: I will try take both of those comments and rerun results. I was really confused about that reordering function thanks for catching this.

td928:
good catch Sasha. I think Max is working on this at the moment. @mbh329 Is there a total population column coming from 2000 pums? If not happy to move the definition `demographic_indicators_denom` into my own script and keep it separate

PR #258, opened 2022-03-29
Export demographics and economics

SashaWeinstein:
# Export the last two EDDT categories 🌻 

We are very close! This PR has new code to export using the 5 accessors we've written together. It also has some changes to the code in those accessors. If you're curious here is [action](https://github.com/NYCPlanning/db-equitable-development-tool/actions/runs/2061537684) that produced the output on DO

## New `collate_save_census` function
This replaces the old `external_review_PUMS.save_PUMS` function. There is a special function for the two census categories because we write separate files for each year-geography. The housing prod/QOL/housing security categories write one file to each geography. 

This new function uses the existing accessor function but doesn't load them from `all_accessors.py`. This is not great design as I wanted one process to load accessors. However given the time constraints I didn't see how to make that happen. So currently we have two classes, `Accessors` in `all_accessors.py` which returns a flat list of indicators used for testing and exporting housing prod/QOL/housing security and then `CensusAccessors` which needs a geography and a year combination to get back an accessor. Maybe the two contexts in which the accessors are called are too different and we won't arrive at one unified process, I'm not sure. We could also use `CensusAccessors` for testing, I know it bother Te that we don't test each year for the census categories. 

### Year parameter
All accessors for demographics and economics now take a year parameter. This is so that they can all be called via ` df = accessor(geography, year)`. For the two census pums accessors this is redundant as they can only take a year parameter of "2000" but it was the best way to have consistency on the export end.
For `decennial_census_001020` the years it can take were changed from `[2000, 2010, 2020]` to ` ["2000", "0812", "1519"]`. This parameter is immediately fed into a mapper dictionary so this change was very easy to make. 

### "pop" variable from census and ACS PUMS re-named
We had run into this problem before where the decennial census has the count of total pop we report on the front-end and the census/ACS PUMS have the total pop we use as a denominator. They were both named "pop" which messes with the merge. We had run into this problem before and I don't remember how we resolved it. My solution here was to rename the pop from ACS/census PUMS to `pop_denom`. If this isn't best it should be pretty straightforward for me or another engineer to change it.

## Miscellaneous changes  

Files, functions and workflows that used to read "PUMS_collate" or "PUMS_export" now are "census_collate" or "census_export". The structure where these two categories need their own code hasn't changed but now we have surveys other than PUMS.


aggregate/all_accessors.py
```diff
@@ -96,30 +96,32 @@
...
+    """Used for testing, exporting QOL/housing security/housing prod"""
+
+    demographics = demographics_accessors
+    economics = economics_accessors
     quality_of_life = QOL_accessors
     housing_production = housing_production_accessors
-    census = census_accessors
     housing_security = housing_security_accessors
-    pums_demographics = demographics_accessors
+    demographics = demographics_accessors
```
td928, line 125: is this a repetition from line 120?  also economics also seems to repeat below

aggregate/clean_aggregated.py
```diff
@@ -71,11 +70,17 @@ def order_affordable(measures, income) -> List:
     for i in income:
         for m in measures:
             rv.append(f"units_affordable_{i}{m}")
-            
+
     return rv
 
 
-def rename_col_housing_security(df: pd.DataFrame, name_mapper: dict, race_mapper: dict, year_mapper: dict, suffix_mapper: dict):
+def rename_col_housing_security(
```
mbh329, line 77: the extra comma at the end is a neat feature in the linter

aggregate/decennial_census/decennial_census_001020.py
```diff
@@ -3,7 +3,7 @@
 from utils.PUMA_helpers import clean_PUMAs
 from internal_review.set_internal_review_file import set_internal_review_files
 
-year_map = {2000: "00", 2010: "10", 2020: "20"}
+year_map = {"2000": "00", "0812": "10", "1519": "20"}
```
mbh329, line 6: cool - thanks for updating this, keep the year string consistent

aggregate/decennial_census/decennial_census_001020.py
```diff
@@ -122,10 +122,10 @@ def create_puma_level_df_by_year(df, year):
 
 
 def decennial_census_001020(
-    geography: str, year: int = 2010, write_to_internal_review=False
+    geography: str, year: str = "2000", write_to_internal_review=False
 ) -> pd.DataFrame:
     assert geography in ["citywide", "borough", "puma"]
-    assert year in [2000, 2010, 2020]
+    assert year in ["2000", "0812", "1519"]
```
mbh329, line 128: The years are changed because they are replacing the 0812/1519 data?

external_review/collate_save_census.py
```diff
@@ -0,0 +1,68 @@
...
+from aggregate.PUMS.pums_2000_economics import pums_2000_economics
+from aggregate.PUMS.pums_0812_1519_demographics import acs_pums_demographics
+from aggregate.PUMS.pums_0812_1519_economics import acs_pums_economics
+
+
+from aggregate.load_aggregated import initialize_dataframe_geo_index
+
+app = typer.Typer()
+
+dec_census_year_mapper = {"1519": 2020, "0812": 2010, "2000": 2000}
```
mbh329, line 18: I think te made a similar comment but I think it makes sense to stay consistent between year as an int vs string

aggregate/all_accessors.py
SashaWeinstein: oh yea good catch lol. deleted the redundant attr assignments

external_review/collate_save_census.py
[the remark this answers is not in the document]
SashaWeinstein: actually that code can just come out, it doesn't do anything

aggregate/decennial_census/decennial_census_001020.py
SashaWeinstein: having the parameter immediately passed to the mapper makes this sort of update really easy. Having clearly defined mappers as python dictionaries is consistently a good pattern for us in this project

td928:
> @td928 I think you're right that functions are more appropriate for our system of returning accessors.
> 
> I'm thinking maybe a single `get_accessors` function that takes a category name and optional arg of year. If the category is housing prod/QOL/housing security then it returns the associated iterable of accessors. If the category is demographics/economics then it uses to year to return the correct iterable of accessors.
> 
> I don't think this process would require partial functions or closures? Do you have an idea of where to implement those?

Hey @SashaWeinstein I think you are right that function are sufficient for our purposes. I also don't think this PR requires that refactoring and just ideas for future refactoring if it happens.

td928 (approved):

PR #263, opened 2022-04-01
Reorder column names with `rate`

mbh329:
quick solution to get proper column names


SashaWeinstein:
Can you rename `col` to `race` or `subgroup` in the for loop? Also could you re-upload the QOL files? All you need to do is change the `export_category.yml` and push to github

SashaWeinstein (approved):
Action looks good

Issue #260, opened 2022-03-31
Add Pop in NYCHA housing to Housing Security output


AmandaDoyle:
Add the following data points to the Housing Security output.
- Total Population (Public Housing and RAD)
     - Count
     - Percent
     - By Race and ethnicity
- 2020 Census Total Pop values, which are in Demographic output.  OSE cannot cross reference tables. 

Logic:

- Total Population (Public Housing and RAD) by race and ethnicity.  Logic is to sum Public Housing White and RAD White, for example
- Denominator to calculate the percent is the 2020 Census Total Pop values, which are in Demographic output

Data source: 
[Equitable Development Data Tool - Displacement Risk Index 2-8-2022 (1).xlsx](https://github.com/NYCPlanning/db-equitable-development-tool/files/8393100/Equitable.Development.Data.Tool.-.Displacement.Risk.Index.2-8-2022.1.xlsx)
This is used for other existing data points.

For reference,the lower box is where these data are being mapped
<img width="687" alt="Screen Shot 2022-03-31 at 3 07 17 PM" src="https://user-images.githubusercontent.com/5611960/161148846-01ec23ba-1d53-4c71-bdbc-5857ecfdd21d.png">

td928:
@AmandaDoyle does the columns naming `rad_pop` and `public_housing_pop` make sense to you for those indicators?

AmandaDoyle:
@td928 there is only one "one column" (i.e. nycha_tenants could be a good name) then nycha_tenants_anh for example.  We need to sum the two columns Public Housing and RAD to get the total number of NYCHA tenants by race and ethnicity.  Ping me if you'd like to hop on a call to talk this through.

SashaWeinstein:
I believe this is the first instance of us referencing a particular data point in two different categories? Curious to see how you do this Te. My first thought is to call the decennial census accessor and take the correct column from the dataframe it returns. Of course there are multiple good ways to do it

td928:
yep this is my thought process too. The census accessor is pretty much what we need apart from the extra pct columns and easy to get rid off. Then hopefully the percent calculation for each race category is straightforward

AmandaDoyle:
@td928 thanks for flagging.  I may be mistaken, but do we have a resource to map 2010 PUMAS to 2020 PUMAS?  If so, we should use that.  If not, then we can do a 1:1 join and leave fields where there is no match blank and flag it to OSE and leave it as a future improvement.

[...]

PR #266, opened 2022-04-01
NYCHA Tenants

td928:
## Overview
addressing #260 from a single spreadsheet and relatively straightforward work 

## `load_clean_nycha_data`
To clean up the puma level NYCHA tenants data.

 the bulk of work takes place where the RAD and public housing population gets summed up by race. 
```python   
for i in range(6):
        nycha_data[f"nycha_tenants{race_labels[i]}_count"] = nycha_data.iloc[:, i] + nycha_data.iloc[:, i + 6] 
```

## `nycha_tenants`
I don't love the design that the join with census data takes place in here. I initially had in the `load_clean_nycha_data` but then realize if the join takes place earlier would mess up the percentage columns for the total population from census. I settled for this current design but open to other suggestions.

## `get_percentage`
to get the percentage of NYCHA population out of total population of the geography

## 
final output has 24 columns each 2 measures (count, pct) * 2 indicators(nycha tenants & pop) * 6 races


aggregate/housing_security/nycha_tenants.py
```diff
@@ -0,0 +1,81 @@
+from aggregate.decennial_census.decennial_census_001020 import decennial_census_001020
+from aggregate.aggregation_helpers import order_aggregated_columns
+import pandas as pd
+from internal_review.set_internal_review_file import set_internal_review_files
+from utils.PUMA_helpers import clean_PUMAs, puma_to_borough
+dcp_pop_races = ["anh", "bnh", "hsp", "onh", "wnh"]
```
mbh329, line 6: can import this dictionary from the `utils/dcp_population_excel_helpers`

mbh329 (commented):
I ran this on my local machine and checked the results, all makes sense to me. I think you just need to add this accessor to `all_accessors.py`

aggregate/housing_security/nycha_tenants.py
td928: which dictionary exactly?

aggregate/housing_security/nycha_tenants.py
```diff
@@ -0,0 +1,81 @@
...
+        "dtype": float
+    }
+    nycha_data = pd.read_excel(**read_excel_arg)
+    nycha_data.rename(columns={"PUMA (2010)": "puma"}, inplace=True)
+    nycha_data.puma = nycha_data.puma.apply(func=clean_PUMAs)
+    nycha_data["borough"] = nycha_data.apply(axis=1, func=puma_to_borough)
+    nycha_data["citywide"] = "citywide"
+    nycha_data.set_index("puma", inplace=True)
+    # calculating the total for each race categories
+    for i in range(6):
```
SashaWeinstein, line 75: I am anxious about iterating over numeric range here. Fragile design that works when run on new source data but could be pointing at wrong columns. Better to iterate through races and then if source data changes the code breaks

aggregate/housing_security/nycha_tenants.py
```diff
@@ -0,0 +1,81 @@
...
+        nycha_data[f"nycha_tenants{race_labels[i]}_count"] = nycha_data.iloc[:, i] + nycha_data.iloc[:, i + 6] 
+
+    return nycha_data.iloc[:, -8:]
+    
+
+def get_percentage(df: pd.DataFrame):
```
SashaWeinstein, line 83: In my opinion it would be simpler to pass `get_percentage` two separate `census20` and `agg` dataframes and do the concat and the calculation here but I think that's an open design choice

SashaWeinstein (changes requested):
I added one commit to clean up an unused call to the decennial census accessor. I also opened an issue to flag the fact that '1519' maps to the 2020 census which is misleading. I made another couple comments too

aggregate/housing_security/nycha_tenants.py
td928: Hey @SashaWeinstein  don't totally disagree about the enumerating over index vs. column name but I need to think about how this might complicate the design. I resorted to this based on the assumptions that 1. this would be one-off for this year and also we need to get this out in a pressing deadline

aggregate/housing_security/nycha_tenants.py
td928: To me there are pros and cons for either design. I like how the current implementation is focused and probably self explanatory. But yeah your design would be less repetitive and cleaner. Any input from @mbh329  or @AmandaDoyle  on this?

td928:
Okay so turns out not a complicated refactor to do it more explicitly at all. @SashaWeinstein I agree it is more readable this way and let me know the new commit addressed your concern. Thanks!

td928:
@SashaWeinstein cool I like what you did with zip. Let me know if you can approve and I will merge it in. Thanks!

SashaWeinstein:
That looks a lot better! I made an additional change to use the python built-in `zip` instead of enumerate because I finally figured out what `zip` does and I wanted to use it.

SashaWeinstein (commented):
Recent commit addresses my biggest concern. Last thing I think is to add this to the housing security accessor list in `all_accessors.py`

PR #271, opened 2022-04-04
Median wages by Occupation by Subgroup

SashaWeinstein:
Easily added with new source data that differentiates between count of workers by occupation-subgroup and median wages of workers by occupation-subgroup. Only by borough internal review files were written as no changes were made that would effect geographies differently. 

One change to script that it looks for a race crosstab on median wages by occupation where it didn't before. 

Can run the github action after this is merged into dev


mbh329 (approved):
Looks good to me

PR #277, opened 2022-04-05
Units Paying Rent Percent and PCT MOE Refactor

td928:
## Overview

Since there is some messiness to the situation due to 1. two indicators share the same columns with `HUPRt` now exists in both `units_affordable` and also `rent_median` so the final collate for output require to eliminate at least one 2. the mistake in the naming the `HUPRt` in the ami source data. There are essentially two steps taken to 

- Removing the `units_renteroccu` columns entirely from the `units_affordable` work entirely
- add the pct and pct_moe columns into the `rent_median` 

a little more details and explanations below for exactly what was done

## `units_affordable`

apart from simply taking away, the biggest change in this part of the refactoring is also removing from the `order_affordable` function to ensure that the final columns ordering no longer has the `units_renteroccu` either. 

## `rent_median`
the only two steps required is first include the `pct` and `pct_moe` in the measurements given to the columns reordering functions to generate the appropriate and removing the `dropna` at the end to ensure they are included. Also two columns the total `units_payingrent` for 1519 are hardcoded to have 100 pct to impute values from ami source file.


aggregate/housing_security/rent_median.py
```diff
@@ -56,14 +56,14 @@ def rent_median(geography: str, write_to_internal_review=False) -> pd.DataFrame:
...
-            measures=["_count", "_count_moe", "_count_cv"],
+            measures=["_count", "_count_moe", "_count_cv", "_pct", "_pct_moe"],
             years=["_0812", "_1519"],
         )
     )
 
     final = pd.concat([final_md, final_hh], axis=1)
 
-    final.dropna(axis=1, how="all", inplace=True)
+    final["units_payingrent_1519_pct"] = 100
```
SashaWeinstein, line 66: Ok looked over how we are setting this column and my understanding is that we have this `EDDT_UnitsAffordablebyAMI_2015-2019.xlsx` file that has this `ROcc2` percent field populated while the ACS PUMS spreadsheets don't? And this pct columns populated with 100's is more correct than the `ROcc2` or `HUPRt` columns in the ACS data?

SashaWeinstein (approved):
Ran the code on my machine and it seems to work. Next step export

aggregate/housing_security/rent_median.py
td928: Hey @SashaWeinstein  didn't meant it to be more confusing but I made a new change and if you want to hear the full explanation for impute the value and keep the column I am happy to talk over a call but it is a little more convoluted

aggregate/housing_security/rent_median.py
```diff
@@ -61,9 +62,17 @@ def rent_median(geography: str, write_to_internal_review=False) -> pd.DataFrame:
...
+    final_hh = final_hh.filter(regex="0812|" + "|".join(race_suffix_mapper.values()))
+
     final = pd.concat([final_md, final_hh], axis=1)
 
     final["units_payingrent_1519_pct"] = 100
+    final["units_payingrent_1519_pct_moe"] = 1
+
+    final.dropna(how="all", axis=1, inplace=True)
+
+    final["units_payingrent_1519_pct_moe"] = np.nan
```
SashaWeinstein, line 74: This block doesn't make sense to me. Why not dropna first and then assign the nan column?

aggregate/housing_security/rent_median.py
SashaWeinstein: Ok if you and Amanda feel good about it I don't want it to hold us up, feel ok approving without understanding exactly what's going on

PR #280, opened 2022-04-07
Change NAN to zero in `nycha_tenants` and `nycha_units`

mbh329:
The source data for `nycha_units` indicator did not provide us with data for PUMA's that did not have nycha units so in the indicator level output there are only 40 pumas (rows of data). In the collate step, the values for the other non reported pumas populate as nan. This PR addresses those issues by filling the nans with zero, pretty simple stuff and used some already established processes (i.e. create empty dataframe of pumas and concat)


td928 (approved):
All look good. Only comment would be if @SashaWeinstein work on the refactoring initializing empty dataframe #218 could be looked at along with this and merged in together.

PR #282, opened 2022-04-07
fix the pre 2010 filter for change in units

td928:
## Overview

My negligence that the pre2010 filter for housing database was never properly added to our main codebase and causing the change in units to be too high. A quick fix to add in the filter and now results should be what is expected.


aggregate/housing_production/change_in_units.py
```diff
@@ -119,6 +120,7 @@ def NYC_PUMA_geographies():
 def clean_jobs(df):
     # DROP INACTIVATE JOBS ACCRODING TO SAM
     df.drop(df.loc[~df.job_inactive.isnull()].index, axis=0, inplace=True)
+    df.drop(df.loc[df["complete_year"] < 2010].index, axis=0, inplace=True)
```
SashaWeinstein, line 123: I prefer to filter on a positive basis, something like 
```python
df = df[df["complete_year"] >= 2010]
```
 is more readable to me personally. But it's also ok the way it's currently written

mbh329 (approved):

SashaWeinstein (approved):
Tests pass and code works on my machine

aggregate/housing_production/change_in_units.py
td928: I agree the double negative gave me trouble as well. But also prefer drop to reassign the value to the same df. But I think the former is greater issue so changed it.

PR #284, opened 2022-04-07
update 2000 demographics with new data for median ages

mbh329:
Erica miscalculated all data points for the median ages in the 2000 pums data so we needed to update the xlsx file in resources with the most recent version. updated the source file in resources and archived the old one. Did a spot check on data and the new outputs align with the updated xlsx but good to get some more eyes on it. No code was changed. Note: the file names are the same but can rename the older version in archive if necessary


AmandaDoyle (approved):
Output looks fine.  Only comment is we may want to append the date to the file names in the archive folder, since there could be multiple version.

mbh329:
@AmandaDoyle appended the date to the archived source xlsx

SashaWeinstein (changes requested):
I think we don't need an archive folder because github stores past versions of each file. In my opinion the new file should just directly replace the old one

mbh329:
@AmandaDoyle said to create an archive folder

SashaWeinstein:
Ok then let's talk about this at standup, my understanding is that github's commit history allows us to avoid storing multiple versions of a file this way

mbh329:
sounds good

td928 (approved):

PR #290, opened 2022-07-26
Auto-format in devcontainer

SashaWeinstein:
Small PR, just requires one reviewer 🍰 
Figured out this fix on the QAQC app first, also applied it here

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -16,7 +16,7 @@ RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/
 # [Optional] Uncomment this section to install additional OS packages.
 # RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
 #     && apt-get -y install --no-install-recommends <your-package-list-here>
-
+RUN apt-get update
 #Install R
 RUN apt-get -y install r-base
 # Install poetry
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -25,3 +25,4 @@ ENV PATH="~/.poetry/bin:$PATH"
 ENV PATH="~/.local/bin:$PATH"
 
 RUN /usr/local/bin/python3 -m pip install -U flake8
+RUN /usr/local/bin/python3 -m pip install -U black
```
mbh329, line 28: Taking a look at the qaqc dockerfile, we install bandit and not flake8. I think it makes sense to stay as consistent as possible

mbh329 (changes requested):

.devcontainer/Dockerfile
mbh329: Rebuilding the dev container and it's taking ages, is this true for you as well @SashaWeinstein

SashaWeinstein: started rebuilding now

mbh329: got it finally

SashaWeinstein: Mine rebuilt pretty quick, meant to look at clock but within 5 minutes I think. Max when was the last time you turned off your computer? 
Also made the requested change from `flake8` to `bandit`

mbh329: yesterday. great will rebuild.

mbh329 (approved):

Issue #293, opened 2022-08-15
Use Version from Data Library Other Than Latest


td928:
Taking another look at how the dataloading works for the open data sources from data library. We haven't handled the cases that versions are not latest. Now the SOE want to make it more explicit on the version being pulled and it needs to look at both on the initial ingestion from data library and also the side of ingestion for each indicator to make sure they are consistent.

PR #294, opened 2022-08-18
source data version control

td928:
#293 multiple reviewers required 🐬 🐬 

## Overview
Stemmed from the request from SOE to hold the source data constant from previous update while also adding new fields to the output which leads to the question: Do we really know which versions of source data we are using in our output? After Max and I figured out which datasets/output are impacted by the open data sources we pulled. This leads me on this mini-expedition to figure out how to keep track of the versions of the few open data sources (for now) in the pipeline. Then I remembered @SashaWeinstein already did this for the facility database, so I went over there and grabbed a lot of the codes off the shelf. 

##`ingest_helpers.py` and `metadata.py`
This works does expand across a few files so might require some patience to review each. But the thread tying everthing together is the new code in `ingest_helpers.py` and `metadata.py`. `metadata.py` is almost verbatim from the facility. I do wonder if we should change it a little to have the dump `metadata.yml` somewhere else but that should be a relatively easy change to make. 

`ingest_helpers.py` was slimmed down a little bit to do just the job. I was debating whether the fetch version from config file is useful for this task and decided we could do without it. So now it takes the version from the `datasets.yml` file and check whether it is downloaded in the `ingest/data_library/dataloading.sh` then would read in the file. 

## datasets and indicator scripts
There are currently only four datasets coming directly from open data. They are all listed in the `datasets.yml` file. Then each of the indicator impacted by those source data were updated with `read_from_s3` functions. Some additional tinkering was required for e.g. historical district where a pandas dataframe is ingested first then converted to geopandas.

## export metadata
using the existing infrastructure for exporting. I modify both the workflow file `export.yml` and also `export_DO.sh` to accomodate the need for the `metadata.yml` exporting. I am depositing this file directly in the root of the `db-eddt` folder on DO.

### DIFF
--- a/.github/workflows/export_category.yml
+++ b/.github/workflows/export_category.yml
@@ -62,6 +62,6 @@ jobs:
         run: |
           poetry run python3 -m external_review.external_review_collate ${{matrix.geography}} ${{matrix.category}}
           ./external_review/export_DO.sh export ${{matrix.geography}} ${{matrix.category}}
-
+          ./external_review/export_DO.sh export_meta
 
 
--- a/aggregate/housing_production/area_within_historic_district.py
+++ b/aggregate/housing_production/area_within_historic_district.py
@@ -4,6 +4,7 @@ from json import load
 import geopandas as gp
 from shapely import wkt
 from utils.PUMA_helpers import puma_to_borough, NYC_PUMA_geographies
+from ingest.ingestion_helpers import read_from_S3
 from internal_review.set_internal_review_file import set_internal_review_files
 
 supported_geographies = ["puma", "borough", "citywide"]
--- a/aggregate/housing_production/area_within_historic_district.py
+++ b/aggregate/housing_production/area_within_historic_district.py
@@ -5,6 +5,7 @@ import geopandas as gp
 from shapely import wkt
 from utils.PUMA_helpers import puma_to_borough, NYC_PUMA_geographies
 from ingest.ingestion_helpers import read_from_S3
+from ingest.data_library.metadata import dump_metadata
 from internal_review.set_internal_review_file import set_internal_review_files
 
 supported_geographies = ["puma", "borough", "citywide"]
@@ -73,7 +74,10 @@ def fraction_PUMA_historic(PUMA, hd):
 
 
 def load_historic_districts_gdf() -> gp.GeoDataFrame:
-    hd = gp.read_file(".library/lpc_historic_district_areas.csv")
+    df = read_from_S3("lpc_historic_district_areas")
+    dump_metadata()
+    hd = gp.GeoDataFrame(df)
+    #hd = gp.read_file(".library/lpc_historic_district_areas.csv")
     hd["the_geom"] = hd["the_geom"].apply(wkt.loads)
     hd.set_geometry(col="the_geom", inplace=True, crs="EPSG:4326")
     hd = hd.explode(column="the_geom")
--- a/aggregate/housing_production/hpd_housing_ny_affordable_housing.py
+++ b/aggregate/housing_production/hpd_housing_ny_affordable_housing.py
@@ -1,7 +1,8 @@
 import pandas as pd
-from ingest.ingestion_helpers import read_from_S3
 from utils.PUMA_helpers import assign_PUMA_col
 from internal_review.set_internal_review_file import set_internal_review_files
+from ingest.ingestion_helpers import read_from_S3
+from ingest.data_library.metadata import dump_metadata
 
 """"We need to download the data, separate the data by construction type (New Construction vs. Preservation)
 as these will be two separate indicators, unit income level, and the various citywide reporting geography 
@@ -21,7 +22,7 @@ unit_income_levels = [
 
 def load_housing_ny():
     """load the HPD Housing NY Units by Building dataset"""
-    df = read_from_S3("hpd_hny_units_by_building.csv",
+    df = read_from_S3("hpd_hny_units_by_building",
                       cols=[
                           "project_id",
                           "project_name",
@@ -42,29 +43,7 @@ def load_housing_ny():
                           "other_income_units",
                       ],
                       )
-
-    df = pd.read_csv(
-        ".library/hpd_hny_units_by_building.csv",
-        usecols=[
-            "project_id",
-            "project_name",
-            "project_start_date",
-            "project_completion_date",
-            "number",
-            "street",
-            "borough",
-            "latitude_(internal)",
-            "longitude_(internal)",
-            "building_completion_date",
-            "reporting_construction_type",
-            "extremely_low_income_units",
-            "very_low_income_units",
-            "low_income_units",
-            "moderate_income_units",
-            "middle_income_units",
-            "other_income_units",
-        ],
-    )
+    dump_metadata()
     df = df.replace(
         {
             "borough": {
@@ -76,6 +55,15 @@ def load_housing_ny():
             }
         }
     )
+    # casting to numeric for calculation
+    num_cols = ["extremely_low_income_units",
+                "very_low_income_units",
+                "low_income_units",
+                "moderate_income_units",
+                "middle_income_units",
+                "other_income_units", ]
+    for c in num_cols:
+        df[c] = pd.to_numeric(df[c])
 
     return df
 
--- a/aggregate/housing_production/hpd_housing_ny_affordable_housing.py
+++ b/aggregate/housing_production/hpd_housing_ny_affordable_housing.py
@@ -1,4 +1,5 @@
 import pandas as pd
+from ingest.ingestion_helpers import read_from_S3
 from utils.PUMA_helpers import assign_PUMA_col
 from internal_review.set_internal_review_file import set_internal_review_files
 
@@ -20,6 +21,28 @@ unit_income_levels = [
 
 def load_housing_ny():
     """load the HPD Housing NY Units by Building dataset"""
+    df = read_from_S3("hpd_hny_units_by_building.csv",
+                      cols=[
+                          "project_id",
+                          "project_name",
+                          "project_start_date",
+                          "project_completion_date",
+                          "number",
+                          "street",
+                          "borough",
+                          "latitude_(internal)",
+                          "longitude_(internal)",
+                          "building_completion_date",
+                          "reporting_construction_type",
+                          "extremely_low_income_units",
+                          "very_low_income_units",
+                          "low_income_units",
+                          "moderate_income_units",
+                          "middle_income_units",
+                          "other_income_units",
+                      ],
+                      )
+
     df = pd.read_csv(
         ".library/hpd_hny_units_by_building.csv",
         usecols=[
@@ -88,7 +111,8 @@ def pivot_and_flatten_index(df, geography):
     for full, abbr in level_mapper.items():
         cols = [c.replace(full, abbr) for c in cols]
 
-    cols = [c.replace("new_construction", "newconstruction") + "_count" for c in cols]
+    cols = [c.replace("new_construction", "newconstruction") +
+            "_count" for c in cols]
 
     df.columns = cols
 
@@ -126,7 +150,8 @@ def citywide_hny_units_con_type(df):
 def borough_hny_units_con_type(df):
 
     results = (
-        df.groupby(["reporting_construction_type", "borough"])[unit_income_levels]
+        df.groupby(["reporting_construction_type", "borough"])[
+            unit_income_levels]
         .sum()
         .reset_index()
     )
@@ -150,7 +175,8 @@ def PUMA_hny_units_con_type(df):
     )
 
     results = (
-        puma_df.groupby(["reporting_construction_type", "puma"])[unit_income_levels]
+        puma_df.groupby(["reporting_construction_type", "puma"])[
+            unit_income_levels]
         .sum()
         .reset_index()
     )
--- a/aggregate/housing_security/income_restricted_units.py
+++ b/aggregate/housing_security/income_restricted_units.py
@@ -8,6 +8,7 @@ from utils.PUMA_helpers import (
 )
 from utils.CD_helpers import nta_to_puma
 from aggregate.load_aggregated import initialize_dataframe_geo_index
+from ingest.ingestion_helpers import read_from_S3
 
 
 def income_restricted_units(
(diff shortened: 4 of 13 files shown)


ingest/ingestion_helpers.py
```diff
@@ -1,7 +1,39 @@
...
+
+def read_from_S3(name: str, cols: list = None) -> pd.DataFrame:
+    read_version = get_dataset_version(name)
+    if os.path.exists(f".library/{name}/{read_version}/{name}.csv"):
+        df = pd.read_csv(
+            f".library/{name}/{read_version}/{name}.csv", dtype=str, index_col=False, usecols=cols
+        )
+    else:
+        return print(f"Cannot find the file version {read_version}. Check version in dataloading for {name}")
+    add_version(dataset=name, version=read_version)
```
SashaWeinstein, line 38: I like that this automatically call `add_version`, I assumed we would call `add_version` in the file that constructs the indicator but this is better

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
```diff
@@ -20,28 +22,28 @@
...
+                          "reporting_construction_type",
+                          "extremely_low_income_units",
+                          "very_low_income_units",
+                          "low_income_units",
+                          "moderate_income_units",
+                          "middle_income_units",
+                          "other_income_units",
+                      ],
+                      )
+    dump_metadata()
```
SashaWeinstein, line 46: This implementation keep overwriting the metadata folder every time it's called. This will work but I feel like there's a cleaner, more intuitive way? Is there somewhere downstream we could call this just once?

SashaWeinstein (commented):
Te this is great stuff, this upgrade is more work than I assumed. I left some initial comments and my next step is run the code and see if it works. 
I will focus more on getting the fellow's PRs merged in this week as there is harder deadline there, so will give more of my attention early next week

ingest/ingestion_helpers.py
```diff
@@ -1,7 +1,39 @@
...
+
+
+def read_from_S3(name: str, cols: list = None) -> pd.DataFrame:
+    read_version = get_dataset_version(name)
+    if os.path.exists(f".library/{name}/{read_version}/{name}.csv"):
+        df = pd.read_csv(
+            f".library/{name}/{read_version}/{name}.csv", dtype=str, index_col=False, usecols=cols
+        )
+    else:
+        return print(f"Cannot find the file version {read_version}. Check version in dataloading for {name}")
```
SashaWeinstein, line 37: I tried to run the code for the first time and this if/else block caught the issue that the version wasn't found. But then it returned `None` and I got a huge stack trace when it hit a bug. Do we want this raise an exception that stops the code? If data isn't found for any indicator we want the whole thing to "fail" right?

SashaWeinstein:
This pipeline is bad what idiot designed this lol

SashaWeinstein (commented):
Ran the code and it works, except for access to subway. Access to subway still uses the old `.library` filepath. Additionally the traffic fatalities has a correct filepath but doesn't write to the metadata file. 

Another thing worth saying is that facDB has the feature where if no version is coded in it will automatically use the latest version. Do we want to implement something similar here?

ingest/data_library/metadata.py
```diff
@@ -0,0 +1,24 @@
...
+class MyDumper(yaml.Dumper):
+    def increase_indent(self, flow=False, indentless=False):
+        return super(MyDumper, self).increase_indent(flow, False)
+
+
+def add_version(dataset: str, version: int):
+    try:
+        version = int(version)
+    except ValueError:
+        version = str(version)
```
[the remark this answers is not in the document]
td928: definitely agree not that proud of what I did there. I think what I am really trying to do there is to default the version to be save as a simple number but resort to a string when needed e.g. devdb's 20Q4.

ingest/ingestion_helpers.py
td928: this should go away once read_from_s3 is using the true s3 workflow

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
td928: I think by metadata folder you meant the metadata file. I thought the process is keep dumping new information into the same metadata.yml file. We might risk having duplicate but shouldn't overwrite the same file.

external_review/export_DO.sh
```diff
@@ -20,7 +20,13 @@ function export_DO_census {
 
 }
 
+function export_metadata (
```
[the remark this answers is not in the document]
td928: don't feel very strongly about this either way so I am gonna change it to do it straight from the workflow yml

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
SashaWeinstein: I do mean metadata file. Did I misunderstand the process? Does this code add to the existing `metdata.yml`? If so I think that's a good approach

aggregate/housing_production/area_within_historic_district.py
```diff
@@ -72,7 +74,9 @@ def fraction_PUMA_historic(PUMA, hd):
 
 
 def load_historic_districts_gdf() -> gp.GeoDataFrame:
-    hd = gp.read_file(".library/lpc_historic_district_areas.csv")
+    df = read_from_S3("lpc_historic_district_areas")
```
mbh329, line 77: Nice to have the `read_from_S3` function to standardize datasets being uploaded via the dataloading step from data library

ingest/data_library/dataloading.sh
```diff
@@ -1,12 +1,6 @@
 #!/bin/bash
 source ingest/data_library/config.sh
 
-import_csv lpc_historic_district_areas
-import_csv doi_evictions
-import_csv dcp_access_subway_sbs
-import_csv dcp_access_ada_subway
-import_csv hpd_hny_units_by_building
-import_csv dcp_housing
 for year in {2010..2020}
```
mbh329, line 4: @td928 with all these files removed, what is the point of having the `dataloading.sh` ?

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
SashaWeinstein: After our conversation I see that the function is named and implemented correctly

ingest/data_library/dataloading.sh
SashaWeinstein: The dot traffic injuries we still import the old way. FacDB also has a combination of dataloading the traditional way and dataloading with `datasets.yml`

SashaWeinstein:
Also I just realized that most files aren't formatted with black. We should address this in this PR or a separate one. Whenever you're ready to think about it

ingest/data_library/dataloading.sh
td928: now it is reduced to only files not following our one file one version convention namely the `dcp_dot_trafficinjuries` which pull all 10 years of data. Also might be in there is the geographies or other datasets we don't necessarily want to consider versioning.

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
```diff
@@ -20,28 +22,29 @@
...
-        ".library/hpd_hny_units_by_building.csv",
-        usecols=[
-            "project_id",
-            "project_name",
-            "project_start_date",
-            "project_completion_date",
-            "number",
-            "street",
-            "borough",
-            "latitude_(internal)",
```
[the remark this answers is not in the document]
mbh329: This behavior isn't caught when running at the borough or citywide level but remains an issue at the PUMA geography

mbh329 (changes requested):
Cast lat/long columns to appropriate datatype

td928:
> Cast lat/long columns to appropriate datatype

@mbh329 let me know if the latest commit fixed this issue for you. Thanks!

mbh329:
The latest commit works on the for `housing_production` at the `puma` level but still an issue with housing_security. Seems to be an issue with evictions

mbh329 (approved):
LGTM

SashaWeinstein (changes requested):
I would prefer we held off on merging this until I reviewed more, is that ok?

mbh329:
Would it be helpful to go over this together with a little code review? @td928 @SashaWeinstein

SashaWeinstein:
The next step for me is to run the new code, can get on a call later in the day

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
```diff
@@ -20,28 +22,29 @@
...
-            "very_low_income_units",
-            "low_income_units",
-            "moderate_income_units",
-            "middle_income_units",
-            "other_income_units",
-        ],
-    )
+    df = read_from_S3("hpd_hny_units_by_building",
+                      "housing_production",
+                      cols=[
```
SashaWeinstein, line 27: Can we return this in a function so we can hard-code it farther down in the file? We made this change consistently in the QAQC app and it made our files much more readable

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
```diff
@@ -53,6 +56,17 @@ def load_housing_ny():
...
     )
+    # casting to numeric for calculation
+    num_cols = ["extremely_low_income_units",
+                "very_low_income_units",
+                "low_income_units",
+                "moderate_income_units",
+                "middle_income_units",
+                "other_income_units",
+                "latitude_(internal)",
+                "longitude_(internal)", ]
```
SashaWeinstein, line 67: Same comment as above

aggregate/housing_security/income_restricted_units.py
```diff
@@ -65,23 +67,24 @@ def income_restricted_units_hpd(
...
+                                   "project_completion_date",
+                                   "number",
+                                   "street",
+                                   "borough",
+                                   "community_board",
+                                   "nta_-_neighborhood_tabulation_area",
+                                   "latitude_(internal)",
+                                   "longitude_(internal)",
+                                   "all_counted_units",
+                               ],
```
SashaWeinstein, line 85: Same comment as above and above

external_review/export_DO.sh
```diff
@@ -8,6 +8,7 @@ function export_DO {
     filename="${category}_${geography_level}.csv"
     SPACES="spaces/edm-publishing/db-eddt/${category}"
     mc cp .staging/$category/$filename $SPACES/$filename
```
SashaWeinstein, line 10: Can we just export the entire `.staging/<category>` folder? That's more in line with how most of our data products work

ingest/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+load_dotenv()
+
+BASE_URL = "https://nyc3.digitaloceanspaces.com/edm-recipes/datasets"
+BASE_PATH = Path(__file__).parent / ".cache"
+
+# if not os.path.isdir(BASE_PATH):
+#     os.makedirs(BASE_PATH, exist_ok=True)
+#     # create .gitignore so that files in this directory aren't tracked
+#     with open(f"{BASE_PATH}/.gitignore", "w") as f:
+#         f.write("*")
```
SashaWeinstein, line 14: Can this code come out? Seems like it can

ingest/data_library/metadata.py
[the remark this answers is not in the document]
SashaWeinstein: Just use `str.isnumeric`, it's exactly what you want https://www.w3schools.com/python/ref_string_isnumeric.asp#:~:text=The%20isnumeric()%20method%20returns,%2C%20and%20the%20%2D%20and%20the%20.

SashaWeinstein (commented):
This is much better! I have some comments, a couple are substantive and the rest aesthetic. 

I'm not so sure about ntext steps thought. I assumed the next step would be to run 3 github actions to test the 3 categories we've changed. However I see that the code currently doesn't next output in a branch folder. It would be easiest to test if we could write to a branch folder. 

Does it make sense to branch off `293-not-latest-from-DO`, make the upgrade to `export_DO` to nest within a folder named after the branch, merge that in, and run a test action? Coding the upgrade to nest within a branch folder would take like 2 seconds

.github/workflows/export_category.yml
```diff
@@ -60,8 +60,8 @@ jobs:
 
       - name: export
         run: |
+          mkdir -p .staging/${{matrix.category}}
```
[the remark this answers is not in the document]
td928: Not sure if it would work if it comes out. Because of how the dump metadata works, the path has to exists before trying to output so I resorted to this. Alternative design would be doing this with Python in the metadata dump function.

external_review/export_DO.sh
td928: sure I thought the initial design was to prevent anything unintended in the final folder which I think is actually a pretty good check even though I was not the author

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
td928: down the file as in which part?

external_review/export_DO.sh
SashaWeinstein: I guess I just disagree. Better to create an output folder, put files in it, and send it up. I'm ok with either way for this data product I just think we usually send the entire `.output` for a reason

.github/workflows/export_category.yml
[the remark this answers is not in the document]
SashaWeinstein: Interesting! Good catch. In that case yea I think we should create the `.staging/<category>` folder in `dump_metadata`. Just copy/paste the two lines from the `collate` function

SashaWeinstein:
Lol Te why did you edit my comment instead of responding

td928:
> Lol Te why did you edit my comment instead of responding

some next level move by me. I guess now we can merge this whole thing in? @SashaWeinstein

SashaWeinstein:
Doing my final review now

.github/workflows/export_category.yml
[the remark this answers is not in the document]
SashaWeinstein: This issue still stands I think

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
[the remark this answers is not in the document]
SashaWeinstein: I assume this still needs to be addressed?

aggregate/housing_production/hpd_housing_ny_affordable_housing.py
SashaWeinstein: Like code up a get_cols function so this can be hard-coded below the import logic. Makes the code more readable

external_review/export_DO.sh
SashaWeinstein: Doesn't seem like I'm getting my way on this

SashaWeinstein (commented):
A couple things left to address. Will open a new PR to format with black so we don't have to do it as part of this PR

SashaWeinstein:
The refactor to move the hardcoding of the columns to separate functions looks good to me. The only thing is that I think the convention is to return `List[str]` instead of `list` https://stackoverflow.com/questions/52623204/how-to-specify-method-return-type-list-of-what-in-python

SashaWeinstein (approved):
This is ready to be merged in! I left the one typing comment but it's not necessary. I was thinking that before we merge we should run the export category workflow and check the output but maybe that's not necessary

PR #295, opened 2022-08-22
Print progress

SashaWeinstein:
Small PR for Te to review 🖨️ 
Common feature in our pipeline to tell the user what's currently running. "calculating" is kind of a silly term here as population does most of the actual calculating but it was the best I could think of. 
I also took out a print statement that I assume was there for testing and debugging


td928 (approved):
lgtm

PR #297, opened 2022-08-25
main and feature branch export

td928:
to test the exporting fully on DO. Modifying the workflow to check if the branch Actions is being run on is main or feature branch.

### DIFF
--- a/.github/workflows/export_category.yml
+++ b/.github/workflows/export_category.yml
@@ -1,9 +1,6 @@
 name: Export Category 
 
 on:
-  push:
-    paths:
-      - ".github/workflows/export_category.yml"
   workflow_dispatch:
     inputs:
       category:
--- a/.github/workflows/export_category.yml
+++ b/.github/workflows/export_category.yml
@@ -61,7 +61,7 @@ jobs:
       - name: export
         run: |
           mkdir -p .staging/${{matrix.category}}
-          poetry run python3 -m external_review.external_review_collate ${{matrix.geography}} ${{matrix.category}}
-          ./external_review/export_DO.sh export ${{matrix.geography}} ${{matrix.category}}
+          poetry run python3 -m external_review.external_review_collate ${{matrix.geography}} ${{matrix.category}} ${{github.ref_name}}
+          ./external_review/export_DO.sh export ${{matrix.geography}} ${{matrix.category}} ${{}}
 
 
--- a/.github/workflows/export_category.yml
+++ b/.github/workflows/export_category.yml
@@ -61,7 +61,7 @@ jobs:
       - name: export
         run: |
           mkdir -p .staging/${{matrix.category}}
-          poetry run python3 -m external_review.external_review_collate ${{matrix.geography}} ${{matrix.category}} ${{github.ref_name}}
-          ./external_review/export_DO.sh export ${{matrix.geography}} ${{matrix.category}} ${{}}
+          poetry run python3 -m external_review.external_review_collate ${{matrix.geography}} ${{matrix.category}} 
+          ./external_review/export_DO.sh export ${{matrix.geography}} ${{matrix.category}} ${{github.ref_name}}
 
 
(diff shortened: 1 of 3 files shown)


SashaWeinstein:
Awesome! Going to test this echoing the spaces variable to the console to check that it's correctly assigned. My only other comment is whether we want to echo all variables passed to the export function in that way

SashaWeinstein:
Another comment here is that [facDB](https://github.com/NYCPlanning/db-facilities/blob/13447e36ce214bbc907c56b6f9d4740ef551e232/facdb.sh#L16) and [pluto](https://github.com/NYCPlanning/db-pluto/blob/0f078657098c6172c28620fc5c8e3eb43d3bfc8d/pluto_build/bin/config.sh#L16) get the branch name in bash instead of the github workflow. Might be nicer to use that approach instead

SashaWeinstein:
Pushed a commit to fix syntax. I also don't remember why we have export census does but I don't know if we need it? Maybe it could come out in this PR

SashaWeinstein:
Maybe export category and census can be two jobs in one build new combined build action?

td928:
> Another comment here is that [facDB](https://github.com/NYCPlanning/db-facilities/blob/13447e36ce214bbc907c56b6f9d4740ef551e232/facdb.sh#L16) and [pluto](https://github.com/NYCPlanning/db-pluto/blob/0f078657098c6172c28620fc5c8e3eb43d3bfc8d/pluto_build/bin/config.sh#L16) get the branch name in bash instead of the github workflow. Might be nicer to use that approach instead

ah this is what I was looking for. Let me know if the most recent commit makes sense.

external_review/export_DO.sh
```diff
@@ -5,8 +5,13 @@ function export_DO {
     echo $2
     geography_level=$1
     category=$2
-    filename="${category}_${geography_level}.csv"
-    SPACES="spaces/edm-publishing/db-eddt/${category}"
+    local filename="${category}_${geography_level}.csv"
+    local branchname=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD)
```
SashaWeinstein, line 9: This looks good to me! I personally want to do more to improve the current export workflow but maybe this isn't the right time for that

SashaWeinstein (approved):
After further conversation with Te we decide to merge this in and wait to implement the bigger upgrade which is scoped out at #298

external_review/export_DO.sh
```diff
@@ -5,8 +5,13 @@ function export_DO {
...
+    if [ "$branchname" = "main" ] ; then
+        SPACES="spaces/edm-publishing/db-eddt/${category}"
+    else
+        SPACES="spaces/edm-publishing/db-eddt/${branchname}/${category}"
+    fi
     mc cp .staging/$category/$filename $SPACES/$filename
     mc cp .staging/$category/metadata.yml $SPACES/metadata.yml
```
SashaWeinstein, line 16: Can we talk again about why not to just export all of `.staging/$category`? I think that would be easier to manage

td928: yeah I think my rationale for it is this eliminate the chances something unexpected ended up DO. It wasn't my design but I like it.

SashaWeinstein: Ok word. I think something unexpected ending up in DO is better than something we output not ending up there. This is why all the other pipelines are written to export everything I think. But it's ok either way I think we should run the action

SashaWeinstein:
Ok I also see that some actions were run from `293-not-latest-from-DO` such as this one https://github.com/NYCPlanning/db-equitable-development-tool/actions/runs/2927282419. I think we should also stop running on pushes to the .yml, that's not a good design. I think that can change can come in this PR

td928:
made another change on the `metadata.py` I think I got to the issue with str vs. int. And also got rid of the run on push in yml. Let me know if looks good on your end to merge. Thanks! @SashaWeinstein

.github/workflows/export_category.yml
```diff
@@ -1,9 +1,6 @@
 name: Export Category 
 
 on:
-  push:
-    paths:
-      - ".github/workflows/export_category.yml"
```
SashaWeinstein, line 6: Awesome

SashaWeinstein (approved):
Great, think we can merge this in. We have this empty folder in DO, we should delete it right https://cloud.digitalocean.com/spaces/edm-publishing?i=266877&path=db-eddt%2F%2F

PR #300, opened 2022-08-30
Make staging directory in dump metadata

SashaWeinstein:
This is what I had in mind to keep the step to create `.staging` out of the workflow yml. I think we should avoid putting logic in the workflow yml files when we can

### DIFF
--- a/external_review/external_review_collate.py
+++ b/external_review/external_review_collate.py
@@ -1,4 +1,5 @@
 """Combine indicators into .csv's to be uploaded to digital ocean"""
+from os import makedirs, path
 import pandas as pd
 import typer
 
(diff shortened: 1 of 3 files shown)


external_review/external_review_collate.py
```diff
@@ -29,9 +28,6 @@ def collate(geography_level, category):
             )
             raise e
     final_df.index.rename(geography_level, inplace=True)
-    folder_path = f".staging/{category}"
```
td928, line 32: I see what you meant now. I am worried that if taking this out from this script would cause issue for categories don't have a meta data dump because they don't use `read_s3` at all?

SashaWeinstein: Oh yea good point. Then it should stay in both places?

Issue #301, opened 2022-08-31
Poetry Installation Deprecation


td928:
t shows an error with [installing poetry](https://github.com/NYCPlanning/db-equitable-development-tool/runs/8116923564?check_suite_focus=true). I did recently come across this issue with gru qaqc now this way to install poetry has deprecated and should be updated probably in many of our data products.

PR #302, opened 2022-09-02
 poetry install update

td928:
#301 

Poetry installation failed in one of the recent GitHub Actions test runs. It them comes to light according to [poetry documentation](https://python-poetry.org/docs/#installation), the installation procedure we being relying on is now deprecated. Along with the new url, the path where poetry gets installed also changed so new path also needs to be added. 

I did [an export](https://github.com/NYCPlanning/db-equitable-development-tool/actions/runs/2972659931) with this feature branch to prove that the new installation worked but feel free to run more extensive tests on this with Actions.

### DIFF
--- a/.github/workflows/export_category.yml
+++ b/.github/workflows/export_category.yml
@@ -40,7 +40,7 @@ jobs:
           sudo apt update
           sudo apt install -y python3-pip python3-distutils
           sudo apt install -y gdal-bin
-          curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3 -
+          curl -sSL https://install.python-poetry.org | python3 -
           curl -O https://dl.min.io/client/mc/release/linux-amd64/mc
           chmod +x mc
           sudo mv ./mc /usr/bin
@@ -48,8 +48,8 @@ jobs:
 
       - name: Poetry Install
         run: |
-          . $HOME/.poetry/env
-          pip install poetry
+          export PATH=$PATH:$HOME/.local/bin
+          
           poetry install      
 
       - name: dataloading
--- a/.github/workflows/export_category.yml
+++ b/.github/workflows/export_category.yml
@@ -49,7 +49,6 @@ jobs:
       - name: Poetry Install
         run: |
           export PATH=$PATH:$HOME/.local/bin
-          
           poetry install      
 
       - name: dataloading
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -21,7 +21,7 @@ RUN apt-get update
 RUN apt-get -y install r-base
 # Install poetry
 RUN curl -sSL https://install.python-poetry.org | python3 -
-ENV PATH="~/.local/bin:$PATH"
+ENV PATH="~/.poetry/bin:$PATH"
 
 RUN /usr/local/bin/python3 -m pip install -U bandit
 RUN /usr/local/bin/python3 -m pip install -U black
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -20,8 +20,7 @@ RUN apt-get update
 #Install R
 RUN apt-get -y install r-base
 # Install poetry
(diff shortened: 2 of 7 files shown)


SashaWeinstein:
Will try devcontainer locally now. Does it make sense to run export category action from this feature branch to check that it works that way too?

SashaWeinstein:
Ran into an issue where poetry wasn't installed on the dev container but it all worked when I pip installed. Adding 
`RUN /usr/local/bin/python3 -m pip install -U poetry` to the bottom of the dockerfile, hopefully this fixes

td928:
> Ran into an issue where poetry wasn't installed on the dev container but it all worked when I pip installed. Adding `RUN /usr/local/bin/python3 -m pip install -U poetry` to the bottom of the dockerfile, hopefully this fixes

I think the issue is this pip installation seems to be deprecated and now moving to [pipx](https://python-poetry.org/docs/#installing-with-pipx). I can try out pipx installation as well. As to testing exporting category, I did a test in my openning comment for the PR. Let me know if you meant something else @SashaWeinstein

SashaWeinstein:
Sorry I did miss the action you linked to in the original PR description. All looks good to me there
And yea pipx does seem like the best installation option as it's simpler than the "official" route linked on the website

td928:
> Sorry I did miss the action you linked to in the original PR description. All looks good to me there And yea pipx does seem like the best installation option as it's simpler than the "official" route linked on the website

But i guess that raises the question whether we want to have two different poetry installation approaches local vs. github actions. Seems not ideal. What exactly is the error you are experiencing with the poetry installer?

SashaWeinstein:
Can we just change both? In my view having `pipx poetry` is more readable than` curl -sSL https://install.python-poetry.org | python3 -`

.github/workflows/export_category.yml
```diff
@@ -40,16 +40,15 @@ jobs:
...
           chmod +x mc
           sudo mv ./mc /usr/bin
           mc config host add spaces $AWS_S3_ENDPOINT $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY --api S3v4
 
       - name: Poetry Install
         run: |
-          . $HOME/.poetry/env
-          pip install poetry
+          export PATH=$PATH:$HOME/.local/bin
           poetry install
```
mbh329, line 52:
```suggestion
          poetry install      
          poetry --version
```

mbh329: Suggesting as the Github Action log suggests this command [here](https://github.com/NYCPlanning/db-equitable-development-tool/runs/8209344747?check_suite_focus=true#step:4:330)

SashaWeinstein: We could have poetry print the version so we can see it in the logs, that seems like a fine suggestion

td928:
> Can we just change both? In my view having `pipx poetry` is more readable than` curl -sSL https://install.python-poetry.org | python3 -`

but it seems like the other way is more dependencies to manage no? My understanding of how pipx is going to work is it first needs to be installed by pip then install pipx like this 
```
python3 -m pip install --user pipx
python3 -m pipx ensurepath
```

SashaWeinstein:
Lol sorry I can't reproduce the error. So between missing the working github action linked in the original description and messing up the run on my local I held this up for no reason. Think it should be good to go as is, open to switching over to pipx for simplicity but this can be merged if we want to move on

SashaWeinstein (approved):
Poetry worked on my devcontainer when I opened it a second time, not sure why I couldn't get it to work the first time.

SashaWeinstein:
> > Can we just change both? In my view having `pipx poetry` is more readable than` curl -sSL https://install.python-poetry.org | python3 -`
> 
> but it seems like the other way is more dependencies to manage no? My understanding of how pipx is going to work is it first needs to be installed by pip then install pipx like this
> 
> ```
> python3 -m pip install --user pipx
> python3 -m pipx ensurepath
> ```

Right that makes sense, no need to install pipx for just poetry

PR #306, opened 2022-09-15
Merge dev into main

SashaWeinstein:
Absolute unit of a PR, should have as many eyes on it as possible ⛰️ ⛰️ ⛰️ 
This was prompted because Tyler confirmed that the new housing security output with the housing lottery data looks good. The new indicators should now be included in the main output. The way to do that is run the new code from main.
I think that the work on dev right now is a fairly stable (for lack of a better term) version of the pipeline and it makes sense to merge it into main.
None of us should check every file like you would a typical pull request, rather we should think conceptually about whether we want this version of the pipeline on the main branch


.github/workflows/export_category.yml
```diff
@@ -0,0 +1,62 @@
...
+      
+    strategy:
+      matrix:
+        geography:
+          - puma
+          - borough
+          - citywide
+        category:
+          - housing_production
+          #- quality_of_life
```
mbh329, line 28: Do we want to keep the version on main with this commented out? I think the default should be that we export all categories, not a big deal to comment out though

SashaWeinstein: This is a good flag, the workflow is messy. It should be changed to either always do all 3 categories or allow the user to select just one. Do you want to implement this upgrade? If not I could

.github/workflows/test_PUMS_aggregation.yml
```diff
@@ -0,0 +1,48 @@
...
+  #   paths:
+  #     - "aggregate/PUMS/**.py"
+  #     - "tests/PUMS/**.py"
+  #     - ".github/workflows/test_PUMS_aggregation.yml"
+  #     - "poetry.lock"
+  #     - "pyproject.toml"
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
mbh329, line 15: Why does this workflow run on the ubuntu-latest image and not specified to 20.02?

mbh329: This question applies to all the test workflows

SashaWeinstein: No good reason, that was the way I set it up. If you want to implement the standardization go for it

mbh329: I made an issue for this. I can open up a feature branch and open a PR into this one

.github/workflows/export_category.yml
SashaWeinstein: This [data library workflow](https://github.com/NYCPlanning/db-data-library/blob/261-Image-Workflow/.github/workflows/update_image.yml) limits the user to a set of choices. That would be a good upgrade here. In the future we could implement an "all" option but that might be hard to achieve right now

mbh329 (commented):
I think this version for the pipeline works for now. I think when we figure out other data updates we will have a better understanding of where we want this repo and workflow to go. once the other PR is approved, I can merge that in and will approve this

SashaWeinstein:
great!

mbh329 (approved):
lgtm

.github/workflows/test_PUMS_aggregation.yml
td928: Is this change going to be made before merging? @mbh329

mbh329: no

Issue #312, opened 2023-03-01
2023 EDDE Update


AmandaDoyle:
Goal: Provide updated EDDE data to OSE by the end of March.
App: https://equitableexplorer.planning.nyc.gov/map/data/district
2023 input data: [here](https://nyco365.sharepoint.com/sites/NYCPLANNING/itd/edm/Shared%20Documents/Forms/AllItems.aspx?RootFolder=%2Fsites%2FNYCPLANNING%2Fitd%2Fedm%2FShared%20Documents%2FPROJECTS%2FDATA%20ENGINEERING%2FEquitable%20Development%20Data%20Tool%2F2023%5FSourceData)

Process:
- Update source data tables [here](https://github.com/NYCPlanning/db-equitable-development-tool/tree/main/resources)
- May want to archive old tables
- If there is no updated source data we are not updating that dataset.  Leave existing data table in place.
- Run the [action](https://github.com/NYCPlanning/db-equitable-development-tool/actions/workflows/export_category.yml) to update the data for an EDDE category
  - based on new source data, only `housing_security` and `quality_of_life` categories are to be updated
- Troubleshoot as needed.  Issues may be schema or underlying data changed.  If this happens and a fix is not simple and obvious let's quickly huddle to figure out how to address problem.
- Updated datasets are pushed up to DO [here](https://cloud.digitalocean.com/spaces/edm-publishing?i=266877&path=db-eddt%2F) and these are what's delivered to OSE.

TODOs since we've reviewed new source data
- [x] request new transportation data to get missing columns per [details below](https://github.com/NYCPlanning/db-equitable-development-tool/issues/312#issuecomment-1452198504)
- [x] archive old resources files
- [x] add new files to repo resources
- [x] do any necessary manual excel file and script edits
- [x] when we get it, process and use new transportation data
- [x] when we get it, use new PUMS data 
- [x] run Export Category action for relevant categories on dev branch
- [x] confirm new data is in `edm-publishing/db-eddt/DEV_BRANCH_NAME/`
- [x] run Export Category action for relevant categories on main branch

damonmcc:
reviewing source data in Sharepoint for completeness
- `EDDE_2023_Updates_transportation.xlsx`
  - incomplete data
  - `access_subway_and_access_ADA.py`
[...]

PR #319, opened 2023-03-23
ingest new transportation data for 2023 update

damonmcc:
## motivations
- part of #312 
- new transportation data was shared by Housing on 3/13

## changes
- add new transportation data
- edit `access_subway_and_access_ADA.py` to process new data
- add `read_from_excel` to `ingestion_helpers.py`
- add `test_ingestion_helpers.py`
- trigger tests on changes to `ingest/**.py`
- don't install R in dev container

## tests
- PR tests passing
- exported categories to edm-publishing on this branch. action runs:
  - [quality_of_life](https://github.com/NYCPlanning/db-equitable-development-tool/actions/runs/4534238486/jobs/7988154512#step:7:36)


fvankrieken (approved):

fvankrieken:
Commenting for posterity - motivation to move this from getting this file from S3 vs local excel is that most of this repo reads in from local files, so mainly for sake of internal consistency.

I still think in a bit, we should maybe move all sources to data library/s3. I get that it's a little annoying if they're all things we grab quite manually as it's just an extra step to put it in s3 and then grab it back, but long term I'd still prefer consistency across projects and getting all source files in edm-recipes in the same manner, rather than having some stored in github

@mbh329 I'm gonna go ahead and merge this

PR #320, opened 2023-03-27
WIP: PUMS data update for 2023 data

fvankrieken:
#312 

A couple big buckets of changes

1. data updates. Mostly added excel files, updated datasets.yml as well
2. Testing/github workflow changes. Largely covered in #321, can mostly ignore for review here
3. Logic around cleaning column names. There was a lot of duplicated logic around cleaning column names relying on specific years, both for things involving the ACS data and not. I further centralized some logic [here](https://github.com/NYCPlanning/db-equitable-development-tool/pull/320/files#diff-6455cf411de83f40eccae07217fdf8706a4b4f04f2798950d4b7acea4a4f90c5) where some of this work had previously been done, then tweaked logic in many files to reflect this. There's also now one hardcoded spot of ACS years [here](https://github.com/NYCPlanning/db-equitable-development-tool/pull/320/files#diff-9591a9057a34b4f40f984b9a9687bfe243f1f2ca735fcc1129c605babaeee6baR8), seemed simplest for now to leave the default args in all the indicators/accessors but make it so that it only needs to be updated in one spot

Most recent actual builds
- https://github.com/NYCPlanning/db-equitable-development-tool/actions/runs/4610789538
- https://github.com/NYCPlanning/db-equitable-development-tool/actions/runs/4610318167

latest s3 outputs
- https://nyc3.digitaloceanspaces.com/edm-publishing/db-eddt/fvk_pums_20230323/2023-04-06

Lastly, theres this file - I really can't find an older equivalent of it. For now I put it and the income adjusted ACS files in the archive folder, just to keep them for posterity
- https://github.com/NYCPlanning/db-equitable-development-tool/pull/320/commits/a839a673061b064ee4e8a42f654220d9d3d663f3#diff-ada96cb96a87db437bffa6a14f1fa4c805d10d179e38a7e782cc3b70404fc030

PR #321, opened 2023-03-30
Build/Github Workflow Improvements

fvankrieken:
Making these changes in separate branch for clarity while reviewing. Sequel to #320, will merge this back into my other feature branch before merging back to main

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,8 +1,8 @@
-# See here for image contents: https://github.com/NYCPlanning/docker-geosupport
+# See here for base image contents: https://github.com/devcontainers/images/tree/main/src/python
+ARG VARIANT="3.9"
+FROM mcr.microsoft.com/devcontainers/python:0-${VARIANT}
 
-# [Choice] Geosupport version
-# ARG VARIANT="latest"
-FROM nycplanning/docker-geosupport:latest
+RUN export DEBIAN_FRONTEND=noninteractive
 
 # [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
 ARG NODE_VERSION="none"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -28,8 +28,8 @@
 	// "forwardPorts": [],
 	// Use 'postCreateCommand' to run commands after the container is created.
 	// "postCreateCommand": "export PATH=$PATH:$HOME/.local/bin && poetry install",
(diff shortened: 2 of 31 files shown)


fvankrieken:
So would welcome any thoughts, things you'd like implemented as well since you just did this build process

fvankrieken:
I think this is pretty set other than updating the readme. Other than getting tests into github workflows that get run automatically, changes workflows around building and exporting as well as the entrypoint scripts for this

1. added `eddt.sh` script as way of accessing functionality. This can be run with 
   - `eddt.sh dataloading` to load data
   - `eddt.sh build` to build. Optional arguments after this are `census` or `category`, which can optionally be followed by more specific args for the python files. So valid examples include
     - `eddt.sh build census` builds all years, econ and dem, all geographies
     - `eddt.sh build census 2000` builds for year 2000, econ and dem, all geographies
     - `eddt.sh build census all demographics citywide`
     - `eddt.sh build category all`
     - `eddt.sh build category housing_production puma`
   - `eddt.sh export`, can optionally specify category (no differentiation between census and category stuff here - so `demographics`, `economics`, `housing_production`, `housing_quality`, `quality_of_life` are the options here, or no arg provided)
 2. changes the workflow yml files for build/export. Right now, there's an `on: push` for one of them because it has no `on: workflow_dispatch` on main so there's no way to trigger it from any branch as far as I can tell. Left as two actions for now. For census, broke into by year just for some granularity with being able to see if a specific one fails. For the other, there's a dropdown to select category but no further granularity (for geography, just does all of them). These functions never had ability to select year, and default to latest. This could be changed as well.

I'll  add above info to readme, but would be happy for review and feedback in the meantime

.devcontainer/Dockerfile
```diff
@@ -1,8 +1,8 @@
-# See here for image contents: https://github.com/NYCPlanning/docker-geosupport
+# See here for base image contents: https://github.com/devcontainers/images/tree/main/src/python
+ARG VARIANT="3.9"
+FROM mcr.microsoft.com/devcontainers/python:0-${VARIANT}
 
-# [Choice] Geosupport version
-# ARG VARIANT="latest"
-FROM nycplanning/docker-geosupport:latest
+RUN export DEBIAN_FRONTEND=noninteractive
```
fvankrieken, line 5: @damonmcc is this from your branch?

damonmcc:
@fvankrieken sounds good. when it's done, may be nice to merge this to main rather than only the PUMS branch

I have a commit I'll push just to clean up the `on:` sections of the test actions

fvankrieken:
@damonmcc gonna merge this into my other feature branch if that sounds good

PR #325, opened 2023-04-19
Fvk update 0812 inflation

fvankrieken:
Part of #312 

Per convo with WS, inflation-adjusted files should be used for 08-12 acs data.

Additionally, I noticed one file which still had 15-19 hardcoded and so wasn't using 17-21

Also addressed - typo by me in the call to my newer column-renaming function:
> There were previously 32 "units_payingrent_" columns, there are now only 2 in the new dataset.

> It looks like only the percent and percent MOE columns made it over and we're missing the unit paying rent's estimate, moe, and CV columns for the total pop and Asian/Black/Hispanic/White households?


fvankrieken:
Added a little bit of a retrospective here

https://github.com/NYCPlanning/db-equitable-development-tool/issues/312#issuecomment-1516372731

mbh329 (approved):
This looks good to me, glad WS and OSE caught the inflation adjusted numbers...maybe in the future we can apply the adjustment in the code ourselves to make things a little easier

damonmcc (approved):

